PR #3, opened 2017-10-19
Adding test pipeline

giorgiosironi:
Only ci stage uncommented, and that instance does not exist yet. But it's a start


giorgiosironi:
Fails as expected.

PR #4, opened 2017-10-24
Setting umask for files written in var/

giorgiosironi:
See https://github.com/elifesciences/journal/commit/8107a2e0f1223cebaa0d9dfba222a6273cfeb962#diff-5f5343d10fae8e62eba1c31ba15d9dc5 for example.
The default is for files to be `rw-r--r--`, but we want them to be `rw-rw-r--` so that the `elife` user and the `www-data` user from the webserver can both use the var/ folder.

Current error:
```
Unable to create the storage directory (/srv/annotations/var/cache/dev/profiler)
```

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -2,6 +2,8 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
+umask(0002);
+
 require __DIR__.'/../vendor/autoload.php';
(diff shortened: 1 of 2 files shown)


giorgiosironi:
The alfred build isn't going to be performed, because we do not have an `annotations--ci` instance yet.

PR #5, opened 2017-10-24
config for ci

giorgiosironi:
Nothing different from prod at the moment


app/config/config_ci.yml
```diff
@@ -0,0 +1,3 @@
+imports:
+  - resource: config_prod.yml
+
```
nlisgo: Nitpick: We only need 1 empty line at the end of the file.

PR #6, opened 2017-10-25
ELPP-3212 Adding build definition

giorgiosironi:


giorgiosironi:
Git has no concept of directory, so what happens with the current `.gitignore` is already that every file inside `build/` is ignored. The `build/.gitkeep` has been force-committed with `git add -f ...` so it is what forces the `build/` folder to be created on new checkouts.

PR #10, opened 2017-11-14
Switch to silex

nlisgo:


src/HypothesisClient/ApiClient/ApiClientInterface.php
```diff
@@ -0,0 +1,15 @@
...
+namespace eLife\HypothesisClient\ApiClient;
+
+use eLife\HypothesisClient\Credentials\CredentialsInterface;
+use eLife\HypothesisClient\HttpClient\HttpClientInterface;
+
+interface ApiClientInterface
+{
+    public function __construct(HttpClientInterface $httpClient, array $headers = []);
+
+    public function setCredentials(CredentialsInterface $credentials);
```
thewilkybarkid: Services shouldn't be stateful. This should be configuration.

src/HypothesisClient/ApiClient/ApiClientInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace eLife\HypothesisClient\ApiClient;
+
+use eLife\HypothesisClient\Credentials\CredentialsInterface;
+use eLife\HypothesisClient\HttpClient\HttpClientInterface;
+
+interface ApiClientInterface
+{
+    public function __construct(HttpClientInterface $httpClient, array $headers = []);
```
thewilkybarkid: No need to put `__construct` on an interface.

src/HypothesisClient/ApiSdk.php
```diff
@@ -0,0 +1,53 @@
...
+            }
+            $this->users = new Users($usersClient, $this->normalizer);
+        }
+
+        return $this->users;
+    }
+
+    public function __call($name, array $args)
+    {
+        throw new BadMethodCallException("Unknown method: {$name}.");
```
thewilkybarkid: Is this really needed?

src/HypothesisClient/Client/ClientInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace eLife\HypothesisClient\Client;
+
+interface ClientInterface
```
thewilkybarkid: If it's a marker interface I'd add a comment.

src/HypothesisClient/Client/Users.php
```diff
@@ -0,0 +1,86 @@
...
+     * @param User $user
+     *
+     * @return PromiseInterface
+     */
+    public function store(User $user) : PromiseInterface
+    {
+        return $this->create($user)
+            ->otherwise(function ($reason) use ($user) {
+                $exception = exception_for($reason);
+                if ($exception instanceof BadResponse && preg_match('/user with username [^\s]+ already exists/', (string) $exception->getResponse()->getBody())) {
```
thewilkybarkid: That's a shame, there's no better way to update/pick out that error?

src/HypothesisClient/Credentials/Credentials.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace eLife\HypothesisClient\Credentials;
+
+use Serializable;
+
+class Credentials implements CredentialsInterface, Serializable
```
thewilkybarkid: Looks like a value object so no need for an interface (the naming here is a giveaway; `Interface` as a suffix is a bad idea).

web/app_local.php
```diff
@@ -0,0 +1,73 @@
+<?php
```
thewilkybarkid: This environment shouldn't be included (it's being removed from other projects since it's not needed/broken).

tests/HypothesisClient/ApiSdkTest.php
```diff
@@ -0,0 +1,80 @@
...
+use PHPUnit_Framework_TestCase;
+
+/**
+ * @covers \eLife\HypothesisClient\ApiSdk
+ */
+final class ApiSdkTest extends PHPUnit_Framework_TestCase
+{
+    /**
+     * @test
+     * @expectedException \BadMethodCallException
```
thewilkybarkid: Recommended way is to the `->expectException()`

tests/Annotations/Command/QueueImportCommandTest.php
```diff
@@ -0,0 +1,171 @@
...
+    {
+        $this->httpClient = $this->getMockBuilder(HttpClient::class)
+            ->setMethods(['send'])
+            ->getMock();
+        $this->apiSdk = new ApiSdk($this->httpClient);
+        $this->application = new Application();
+        $this->limit = $this->createMock(Limit::class);
+        $this->logger = $this->getMockBuilder(LoggerInterface::class)
+            ->setMethods(['emergency', 'alert', 'critical', 'error', 'warning', 'notice', 'info', 'debug', 'log'])
+            ->getMock();
```
thewilkybarkid: Better to use something like [BufferingLogger](https://github.com/symfony/symfony/blob/5129c4cf7e294b1a5ea30d6fec6e89b75396dcd2/src/Symfony/Component/Debug/BufferingLogger.php) rather than constructing mocks of external code manually.

src/HypothesisClient/Model/User.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace eLife\HypothesisClient\Model;
+
+use Assert\Assert;
+
+final class User implements ModelInterface, CanBeValidated
+{
+    use ModelTrait;
+
+    protected $username;
```
thewilkybarkid: `private`?

src/HypothesisClient/Model/User.php
```diff
@@ -0,0 +1,76 @@
...
+        string $email,
+        string $displayName
+    ) {
+        $this->username = $username;
+        $this->email = $email;
+        $this->displayName = $displayName;
+        $this->validate();
+    }
+
+    public function getId() : string
```
thewilkybarkid: ID or username?

src/HypothesisClient/Model/ModelTrait.php
```diff
@@ -0,0 +1,18 @@
...
+trait ModelTrait
+{
+    protected $new = false;
+
+    public function isNew(): bool
+    {
+        return $this->new;
+    }
+
+    public function setNew()
```
thewilkybarkid: Think a custom serializer to add it in the constructor rather than making it stateful would be a better approach.

src/HypothesisClient/Model/CanBeValidated.php
```diff
@@ -0,0 +1,15 @@
...
+use InvalidArgumentException;
+
+interface CanBeValidated
+{
+    /**
+     * @throws InvalidArgumentException
+     *
+     * @return bool
+     */
+    public function validate() : bool;
```
thewilkybarkid: This appears not to be used anywhere except inside the constructor of `User`, so could never be `false`? (Post-initialisation validation is a bad pattern anyway.)

src/HypothesisClient/Client/Users.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Store the user by create first then, if user already detected, modify.
+     *
+     * @param User $user
+     *
+     * @return PromiseInterface
+     */
+    public function store(User $user) : PromiseInterface
```
thewilkybarkid: `upsert` maybe?

src/HypothesisClient/HttpClient/HttpClientInterface.php
```diff
@@ -1,11 +1,11 @@
 <?php
 
-namespace eLife\HypothesisClient;
+namespace eLife\HypothesisClient\HttpClient;
 
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Http\Message\RequestInterface;
 
-interface HttpClient
+interface HttpClientInterface
```
thewilkybarkid: I'll look at extracting this into a separate library, as it could be shared between our API client and the Hypothesis one (so a library for JSON APIs).

composer.json
```diff
@@ -1,82 +1,48 @@
...
-        "guzzlehttp/guzzle": "^6.3",
+        "giorgiosironi/eris": "^0.9.0",
         "mtdowling/jmespath.php": "^2.4",
-        "phpspec/phpspec": "^4.0",
-        "phpspec/prophecy": "^1.7",
         "phpunit/phpunit": "^5.7",
-        "sensio/generator-bundle": "^3.1",
-        "symfony/phpunit-bridge": "^3.3"
+        "silex/web-profiler": "^2.0",
+        "symfony/browser-kit": "^3.3"
```
giorgiosironi: `browser-kit` seems strange here, as there should be no HTML pages to parse

giorgiosironi:
Trying to get this to build in the Vagrant VM, `/ping` seems not to be responding despite the provider being configured.

giorgiosironi:
Weird:
```
PHP message: PHP Fatal error:  Uncaught Error: Call to undefined function eLife\Bus\Limit\pcntl_signal() in /srv/annotations/vendor/elife/bus-sdk/src/Limit/SignalsLimit.php:19
```

src/Annotations/Kernel.php
```diff
@@ -0,0 +1,331 @@
...
+            return new AnnotationsController($app['logger'], $app['cache']);
+        };
+
+        $app->register(new ConsoleServiceProvider(), [
+            'console.name' => 'Annotations console',
+            'console.version' => '0.1.0',
+            'console.project_directory' => self::ROOT,
+        ]);
+
+        $app->register(new AnnotationsServiceProvider(), [
```
giorgiosironi: this seems to create all the objects listed, is this expected?

.travis.yml
```diff
@@ -2,36 +2,45 @@
 language: php
 
 php:
+  - 7.0
   - 7.1
```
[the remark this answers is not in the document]
nlisgo: No additional benefit, I had it running before integrating it with Jenkins.

src/HypothesisClient/ApiClient/ApiClientInterface.php
nlisgo: Some of the queries to the Hypothesis API can be done without credentials. Listing public annotations is one of them.

.travis.yml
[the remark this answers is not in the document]
nlisgo: It also allows me to integrate easily with coveralls, but if that is possible with Jenkins to I can drop this.

src/HypothesisClient/Client/Users.php
nlisgo: I'm considering just reacting to the `BadResponse`. Potentially, it might let one through that we could have also expected to fail for the patch request but the check would be less fragile. The most likely cause for the `Bad Request` will be that the username already exists.

composer.json
nlisgo: We need it for the `/ping` test. It's just a `--dev` dependency though.

giorgiosironi: Possibly came from https://github.com/elifesciences/annotations/blob/develop/tests/WebTestCase.php#L5? I'll do an experiment

giorgiosironi: Ah, it's for `AppKernelTest`, I thought Silex wouldn't force it, but it does

bin/console
```diff
@@ -1,27 +1,16 @@
...
 use Symfony\Component\Console\Input\ArgvInput;
-use Symfony\Component\Debug\Debug;
 
-// if you don't want to setup permissions the proper way, just uncomment the following PHP line
-// read https://symfony.com/doc/current/setup.html#checking-symfony-application-configuration-and-setup
-// for more information
-//umask(0000);
+require_once __DIR__.'/../vendor/autoload.php';
 
+umask(002);
```
giorgiosironi: it doesn't make a difference for this particular number, but umasks are octal so should be `0002` for clarity (leading `0` indicates it's an octal number, then `002` is `rw-rw-r--`)

bin/console
```diff
@@ -1,27 +1,16 @@
...
 set_time_limit(0);
 
-require __DIR__.'/../vendor/autoload.php';
-
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], getenv('SYMFONY_ENV') ?: 'dev');
```
giorgiosironi: we haven't found a way to provide an env variable to long-running processes, so for now `--env` is the best way

smoke_tests.sh
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env bash
-. /opt/smoke.sh/smoke.sh
+set -ex
 
-bin/console --version --env=$ENVIRONMENT_NAME
```
giorgiosironi: the `curl` command is equivalent to the previous smoke test, but doesn't check the body (it's not a massively useful smoke testing framework)
the console command should still be there to check the cli bootstrap works, if `ENVIRONMENT_NAME` is a problem for using the script locally, you can add a default with `"${ENVIRONMENT_NAME:-dev}`

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+use Symfony\Component\HttpKernel\TerminableInterface;
+
+final class AppKernel implements ContainerInterface, HttpKernelInterface, TerminableInterface
+{
+    private $app;
+
+    public function __construct(string $environment = 'dev')
+    {
+        $configFile = __DIR__.'/../../config.php';
+        $config = array_merge(file_exists($configFile) ? require $configFile : [], require __DIR__."/../../config/{$environment}.php");
```
giorgiosironi: as it's written now `config.php`, but `config/$environment.php` is mandatory. Either a `file_exists` is needed on the second `require` too, or empty files are needed for `config/end2end.php` and `config/continuumtest.php` (identical to `config/prod.php`)

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+
+        $this->app = new Application([
+            'debug' => $config['debug'] ?? false,
+            'ttl' => $config['ttl'] ?? 300,
+            'logging.path' => $config['logging.path'] ?? __DIR__.'/../../var/logs',
```
giorgiosironi: should this be `$config['logging.path']` or `$config['logging']['path']`? Given PHP is the format of the configuration file, I think the latter is more idiomatic

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+            ]);
+        }
+
+        // General cache.
+        $this->app['cache'] = function () {
+            return new FilesystemCache(__DIR__.'/../../var/cache');
+        };
+
+        $this->app['logger'] = function (Application $app) {
+            $factory = new LoggingFactory($app['logging.path'], 'annotations-api', $app['logging.level']);
```
giorgiosironi: would just say `annotations` as 2nd argument, since this logger also covers the cli processes

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+        };
+
+        $this->app['monitoring'] = function () {
+            return new Monitoring();
+        };
+
+        /*
+         * @internal
+         */
+        $this->app['limit._memory'] = function (Application $app) {
```
giorgiosironi: later we will extract a provider for this in `bus-sdk`, but not now

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+                $app['logger']
+            );
+        };
+
+        $this->app['hypothesis.guzzle'] = function (Application $app) {
+            // Create default HandlerStack
+            $stack = HandlerStack::create();
+            $logger = $app['logger'];
+            if ($app['debug']) {
+                $stack->push(
```
giorgiosironi: If it's in debug mode, I think we can go the whole mile and use https://github.com/guzzle/psr7#function-str to log the whole request
Also, the same is useful for the response (`GuzzleHttp\Middleware::mapResponse()`)

giorgiosironi: Alternatively, we can log in `$notifyingHttpClient->addRequestListener()` 20 lines from here

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+
+        $this->app->register(new QueueCommandsProvider(), [
+            'sqs.queue_message_type' => $this->app['aws']['queue_message_default_type'],
+            'sqs.queue_name' => $this->app['aws']['queue_name'],
+            'sqs.region' => $this->app['aws']['region'],
+        ]);
+
+        // Cache.
+        if ($this->app['ttl'] > 0) {
+            $this->app->after([$this, 'cache'], 3);
```
giorgiosironi: `3`?

src/Annotations/Command/QueueImportCommand.php
```diff
@@ -0,0 +1,144 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Throwable;
+
+final class QueueImportCommand extends Command
```
giorgiosironi: this was the candidate for `bus-sdk-php` later, I guess

src/Annotations/Command/QueuePushCommand.php
```diff
@@ -0,0 +1,50 @@
...
+use eLife\Bus\Queue\InternalSqsMessage;
+use eLife\Bus\Queue\WatchableQueue;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+final class QueuePushCommand extends Command
```
giorgiosironi: also this a candidate for `bus-sdk-php`, but later

tests/Annotations/Provider/QueueCommandsProviderTest.php
```diff
@@ -0,0 +1,95 @@
...
+use Knp\Provider\ConsoleServiceProvider;
+use LogicException;
+use PHPUnit_Framework_TestCase;
+use Psr\Log\LoggerInterface;
+use Silex\Application;
+
+/**
+ * @covers \eLife\Annotations\Provider\QueueCommandsProvider
+ */
+final class QueueCommandsProviderTest extends PHPUnit_Framework_TestCase
```
giorgiosironi: I don't think this test is very predictive of bugs, since Providers are factories and stubbing all of its dependencies takes away all their responsibility, to put together objects. It could have more value if it used the real provider registered in the Kernel. Otherwise it could be dropped for ease of maintenance, and all the construction logic would be tested at an higher level by the build (which will run commands for example).

web/app_dev.php
```diff
@@ -1,28 +1,15 @@
 <?php
```
giorgiosironi: `web/app_end2end.php` and `web/app_continuumtest.php` are missing, if in doubt they should be identical to `web/app_prod.php`

web/app_prod.php
```diff
@@ -0,0 +1,15 @@
...
+require_once __DIR__.'/../vendor/autoload.php';
+
+$app = new AppKernel('prod');
+
+$request = Request::createFromGlobals();
+
+$response = $app->handle($request);
+$response->send();
+
+$app->terminate($request, $response);
```
giorgiosironi: got to the last line

config.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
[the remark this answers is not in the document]
nlisgo: It's gone

src/Annotations/AppKernel.php
nlisgo: I have added `config/end2end.php` and `config/continuumtest.php`

src/Annotations/Command/QueueImportCommand.php
nlisgo: Yes

src/Annotations/Command/QueuePushCommand.php
nlisgo: Yes

web/app_dev.php
nlisgo: Have added

src/HypothesisClient/Client/Users.php
```diff
@@ -0,0 +1,91 @@
...
+                 * already exists. Because this can only be determined by the
+                 * text in the response, it is considered a bit fragile. Until
+                 * Hypothesis set an error code in their response we will treat
+                 * all BadResponse's as if they are for a known username and
+                 * attempt an update request.
+                 */
+                if ($exception instanceof BadResponse) {
+                    return $this->update($user);
+                } else {
+                    throw $exception;
```
thewilkybarkid: Best not to force failures to be an exception, drop the `exception_for()` above and do `return rejection_for($reason)` here. The `else` is not needed either.

src/HypothesisClient/ApiClient/ApiClient.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace eLife\HypothesisClient\ApiClient;
+
+use eLife\HypothesisClient\Credentials\Credentials;
+
+interface ApiClient
+{
+    public function getCredentials() : Credentials;
```
thewilkybarkid: Is this needed? Never seems to be used externally?

src/HypothesisClient/ApiClient/ApiClientTrait.php
```diff
@@ -1,22 +1,38 @@
...
+use eLife\HypothesisClient\Credentials\Credentials;
+use eLife\HypothesisClient\HttpClient\HttpClient;
 use eLife\HypothesisClient\HttpClient\UserAgentPrependingHttpClient;
 use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Psr7\Request;
-use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UriInterface;
 
-trait ApiClient
+trait ApiClientTrait
```
thewilkybarkid: Since this is currently sharing the same name as something else, think about what the trait is for.

src/HypothesisClient/ApiClient/ApiClientTrait.php
```diff
@@ -36,24 +52,36 @@ public function __construct(HttpClient $httpClient, array $headers = [])
...
 
         return $this->httpClient->send($request);
     }
+
+    final protected function patchRequest(
+        UriInterface $uri,
+        array $headers,
+        string $content
+    ) : PromiseInterface {
+        $headers = array_merge($this->headers, $headers);
```
thewilkybarkid: Maybe have a `createRequest()` method to avoid duplicating this?

src/HypothesisClient/ApiSdk.php
```diff
@@ -0,0 +1,44 @@
...
+     * @var Users
+     */
+    private $users;
+
+    public function __construct(HttpClient $httpClient, Credentials $credentials = null)
+    {
+        $this->httpClient = $httpClient;
+        $this->credentials = $credentials;
+        $this->normalizer = new ObjectNormalizer(null, new CamelCaseToSnakeCaseNameConverter());
+        $this->users();
```
thewilkybarkid: Is this meant?

src/HypothesisClient/Client/Client.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace eLife\HypothesisClient\Client;
+
+/**
+ * Interface Client.
```
thewilkybarkid: 🤔

src/HypothesisClient/Model/ModelTrait.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace eLife\HypothesisClient\Model;
+
+trait ModelTrait
```
thewilkybarkid: `CanBeNew`?

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+            $this->app->register(new ServiceControllerServiceProvider());
+            $this->app->register(new TwigServiceProvider());
+            $this->app->register(new WebProfilerServiceProvider(), [
+                'profiler.cache_dir' => __DIR__.'/../../var/cache/profiler',
+                'profiler.mount_prefix' => '/_profiler',
+            ]);
+        }
+
+        // General cache.
+        $this->app['cache'] = function () {
```
thewilkybarkid: Where's this used?

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,292 @@
...
+
+        return $this->app[$id];
+    }
+
+    public function has($id) : bool
+    {
+        return isset($this->app[$id]);
+    }
+
+    public function getApp() : Application
```
thewilkybarkid: This doesn't look needed. Only appears to be used in tests, where `AppKernel` itself can be used.

composer.json
```diff
@@ -1,82 +1,48 @@
...
-        "sensio/distribution-bundle": "^5.0",
-        "sensio/framework-extra-bundle": "^5.0",
-        "symfony/monolog-bundle": "^3.1",
-        "symfony/polyfill-apcu": "^1.5",
-        "symfony/serializer": "^3.3",
-        "symfony/symfony": "^3.3",
-        "twig/twig": "^2.4"
+        "php": "^7.0",
+        "beberlei/assert": "^2.7",
+        "doctrine/annotations": "1.4.0",
```
thewilkybarkid: Is this used?

composer.json
```diff
@@ -1,82 +1,48 @@
...
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget"
-        ],
-        "post-install-cmd": [
-            "@symfony-scripts"
-        ],
-        "post-update-cmd": [
-            "@symfony-scripts"
-        ]
-    },
+    "minimum-stability": "dev",
```
thewilkybarkid: Always a little suspicious in a project, what needs this?

src/Annotations/Command/QueueImportCommand.php
```diff
@@ -0,0 +1,144 @@
...
+    private $queue;
+    private $limit;
+
+    public function __construct(
+        ApiSdk $sdk,
+        WatchableQueue $queue,
+        LoggerInterface $logger,
+        Monitoring $monitoring,
+        Limit $limit,
+        $supports = []
```
thewilkybarkid: Not sure what this is for?

nlisgo: This will have more relevance when the command in adopted into `bus-sdk`. For a project such as `annotations` we only want to import profiles into the queue. But if left blank then all the possible content that could be queued will be imported.

src/Annotations/AppKernel.php
nlisgo: Nothing at the moment, I'll remove.

src/Annotations/AppKernel.php
nlisgo: I'll remove.

src/Annotations/Command/QueueImportCommand.php
thewilkybarkid: Only looks to work with profiles though? (`importProfiles`)

src/HypothesisClient/ApiClient/ApiClient.php
nlisgo: No. Will remove.

src/HypothesisClient/ApiClient/ApiClientTrait.php
nlisgo: 👍

composer.json
giorgiosironi: Years ago this was necessary to install private libraries which hadn't a defined release version, but only `master`. Perhaps `api-sdk` and such don't need it?

src/Annotations/Command/QueueImportCommand.php
giorgiosironi: I agree it should eventually go upstream in `bus-sdk`, no opinion on whether to leave the parameter in now or to drop it to use the constant.

src/HypothesisClient/ApiClient/ApiClientTrait.php
nlisgo: I'm just going to get rid of the marker interface.

src/HypothesisClient/ApiSdk.php
nlisgo: I'll refactor

src/HypothesisClient/Model/ModelTrait.php
nlisgo: 👍

composer.json
nlisgo: Nope

composer.json
nlisgo: this can go. :composer-noob:

src/Annotations/Command/QueueImportCommand.php
nlisgo: I can drop it now and introduce when I add the command to `bus-sdk`.

src/HypothesisClient/Client/Client.php
nlisgo: I'm just going to remove the marker interface

web/app_prod.php
nlisgo: :1st_place_medal:

src/Annotations/AppKernel.php
nlisgo: Should we also add it for the calls to the elife api for the profiles requests?

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,281 @@
...
+            $stack = HandlerStack::create();
+            $logger = $app['logger'];
+            if ($app['debug']) {
+                $stack->push(
+                    Middleware::mapRequest(function ($request) use ($logger) {
+                        $logger->debug("Request performed in Guzzle Middleware: {$request->getUri()}");
+
+                        return $request;
+                    })
+                );
```
nlisgo: @giorgiosironi do you also want me to expand what we are logging here?

giorgiosironi: best-effort, we can easily tune this later and the current level (`debug`) poses no risk

src/Annotations/AppKernel.php
```diff
@@ -0,0 +1,290 @@
...
+        ]);
+
+        $this->app->register(new ApiProblemProvider());
+        $this->app->register(new PingControllerProvider());
+
+        if ($this->app['debug']) {
+            $this->app->register(new VarDumperServiceProvider());
+            $this->app->register(new HttpFragmentServiceProvider());
+            $this->app->register(new ServiceControllerServiceProvider());
+            $this->app->register(new TwigServiceProvider());
```
giorgiosironi: Twig is probably not needed?

src/Annotations/AppKernel.php
giorgiosironi: At the `debug` level too, could be useful; not a blocker though

PR #11, opened 2017-11-15
Lazy creation of basic objects

giorgiosironi:
When objects are requested from $app and passed to `AnnotationsServiceProvider`, they are instantiated.

This is a bit wasteful, but more importantly some of them can only be built from the cli and not in php-fpm, as our configuration disables the `pcntl_*` functions for security reasons.

We should probably go for a similar pattern for the other objects as well, but they don't break the application on our servers.

### DIFF
--- a/src/Annotations/Kernel.php
+++ b/src/Annotations/Kernel.php
@@ -247,8 +247,6 @@ final class Kernel implements MinimalKernel
 
         $app->register(new AnnotationsServiceProvider(), [
             'annotations.hypothesis.sdk' => $app['hypthesis.sdk'],
-            'annotations.limit.import' => $app->protect($app['limit.interactive']),
-            'annotations.limit.watch' => $app->protect($app['limit.long_running']),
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Builds locally, but not on `ci`, finding out why.
#worksonmymachine

giorgiosironi:
```
2017/11/15 12:18:04 [error] 1200#1200: *7 FastCGI sent in stderr: "Unable to open primary script: /srv/annotations/web/app.php (No such file or directory)" while reading response header from upstream, client: 127.0.0.1, server: localhost, request: "GET /ping HTTP/1.1", upstream: "fastcgi://unix:/var/php-fpm.sock:", host: "localhost"
```

giorgiosironi:
Made some tweaks to the formula, this is now green and should make theother PR it is merge into green.

thewilkybarkid:
Good point. In fact, there's no need to pass any service in that way.

PR #18, opened 2017-11-22
Improve testing for profile email address (found and not found)

nlisgo:
Have also improved testing for other log messages

### DIFF
--- a/tests/Annotations/Command/QueueWatchCommandTest.php
+++ b/tests/Annotations/Command/QueueWatchCommandTest.php
@@ -21,7 +21,6 @@ use eLife\Logging\Monitoring;
 use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Psr7\Request;
 use PHPUnit_Framework_TestCase;
-use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Tester\CommandTester;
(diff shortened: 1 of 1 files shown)


tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -121,6 +124,20 @@ public function it_will_process_an_item_in_the_queue(QueueItem $item, Profile $p
         $this->assertEquals(1, $this->queue->count());
         $this->commandTesterExecute();
         $this->assertEquals(0, $this->queue->count());
+        $actual_logs = $this->logger->cleanLogs();
```
giorgiosironi: nice way to capture everything that is logged

tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -142,6 +162,36 @@ public function providerProfiles()
...
+        yield 'with single email' => [
+            new InternalSqsMessage('profile', 'username'),
+            new Profile('username', new PersonDetails('PreferredName', 'IndexName'), new EmptySequence(), new ArraySequence(['username@email.com'])),
+            [
+                'username' => 'username',
+                'email' => 'username@email.com',
+                'display_name' => 'PreferredName',
+            ],
+            [],
+            [
```
giorgiosironi: While the whitelist seems very precise in specifying what has to be logged, the blacklist is more tricky as the specified entry could very well be logged with a different level, or with a very slightly different message (e.g. without the final `.`)

Perhaps substitute the blacklist with an assertion that *all* log messages must be in the whitelist, otherwise they are not accounted for and fail the test?

giorgiosironi: Design-wise, in cases where we need this level of introspection for what is happening an Observer pattern can come up, but let's not make it more complex. I suspect the logic in the command will soon grow and something will be extracted from it.

nlisgo: Having reviewed these blacklist checks, they add no value so I have removed them.

PR #20, opened 2017-11-24
Fail fast on generic Bad Response on POST /users

giorgiosironi:
Encountered this yesterday by chance.

This is the response in case of an already existing user:
```
HTTP/1.1 400 Bad Request\r\n
Date: Fri, 24 Nov 2017 11:34:19 GMT\r\n
Content-Type: application/json; charset=UTF-8\r\n
Content-Length: 155\r\n
...

{"status": "failure", "reason": "user with email address m8nl1gze@hypothesis.elifesciences.org already exists, user with username m8nl1gze already exists"}
```

We have decided previously not to rely on the body error message as it's brittle and it changes depending on the input. We can still check the status code as a first filter however: 405 and other 40X are still considered a `BadResponse` but they should fail immediately as they are outside of the scope of an existing user.

### DIFF
--- a/src/HypothesisClient/Client/Users.php
+++ b/src/HypothesisClient/Client/Users.php
@@ -53,7 +53,9 @@ final class Users
                  * attempt an update request.
                  */
                 if ($reason instanceof BadResponse) {
-                    return $this->update($user);
+                    if ($reason->getResponse()->getStatusCode() == 400) {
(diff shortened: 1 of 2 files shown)


giorgiosironi:
Have to leave this closed, or merged, due to Jenkins debugging in progress. Feel free to reopen and merge.

PR #27, opened 2017-11-27
Stub AWS SQS in only integration test that requires it

giorgiosironi:
Wouldn't move this setup to the base `WebTestCase` as other tests should
not need it, being focused on single classes.

### DIFF
--- a/src/Annotations/AppKernel.php
+++ b/src/Annotations/AppKernel.php
@@ -283,4 +283,14 @@ final class AppKernel implements ContainerInterface, HttpKernelInterface, Termin
     {
         return isset($this->app[$id]);
     }
+
+    /**
+     * Use only in integration tests.
+     *
+     * @internal
+     */
+    public function override($id, callable $factory)
+    {
+        $this->app[$id] = $factory;
+    }
 }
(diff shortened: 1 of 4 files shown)


src/Annotations/AppKernel.php
```diff
@@ -283,4 +283,14 @@ public function has($id) : bool
     {
         return isset($this->app[$id]);
     }
+
+    /**
+     * Use only in integration tests.
+     *
+     * @internal
+     */
+    public function override($id, callable $factory)
```
thewilkybarkid: Seems like something that should be done when constructing the App in the CI environment?

giorgiosironi: More like a `test` environment like in journal, because the console and the smoke tests should still use a `SqsWatchableQueue` (whether it is backed by goaws or the real AWS)

nlisgo: Since we have the `$app['aws']['stub']` flag, we don't need this `override` method.

giorgiosironi: Right, removed

PR #29, opened 2017-11-29
ELPP-3304 Use access control on emailAddresses

giorgiosironi:
Annotations will only see public email addresses at the moment, so we don't have to filter.

However, they are now wrapped into an `AccessControl` object that specifies their access (`public` or `restricted`).

### DIFF
--- a/src/Annotations/Command/QueueWatchCommand.php
+++ b/src/Annotations/Command/QueueWatchCommand.php
@@ -2,6 +2,7 @@
 
 namespace eLife\Annotations\Command;
 
+use eLife\ApiSdk\Model\AccessControl;
 use eLife\ApiSdk\Model\Profile;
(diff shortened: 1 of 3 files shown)


src/Annotations/Command/QueueWatchCommand.php
```diff
@@ -43,7 +44,9 @@ protected function process(InputInterface $input, QueueItem $item, $entity = nul
     {
         if ($entity instanceof Profile) {
             $id = $entity->getIdentifier()->getId();
-            $emails = $entity->getEmailAddresses();
+            $emails = $entity->getEmailAddresses()->map(function (AccessControl $accessControl) {
```
thewilkybarkid: I'd not rely on the client having only requested public data, and add a `filter()`. Avoids future mistakes.

PR #30, opened 2017-11-29
ELPP-3304 Follow up access control change

thewilkybarkid:
Follow up to #29 now the BC layer has been removed.


giorgiosironi:
Let's go with this.

PR #31, opened 2017-12-04
ELPP-3321 Container(s)

giorgiosironi:
Really experimental version of containers that could be used to develop this project on a laptop without installing PHP or other dependencies. See notes intertwined with the files.

## Features

- build images locally for `fpm` and `cli` (long-running process)
- run `composer` inside a container (experimental)
- run `phpunit` inside a container (stable), including by filtering a single test
- run a full environment with `nginx`, `fpm`, `cli`, `goaws` and `logger`

## Checklist

To be done:
- [x] `cli` container
- [x] `goaws` container to be brought in (it's a dependency for `cli`)
- [x] make `containerized` configuration go away
- [ ] avoiding `web/` folder in nginx as there are no files to serve from there. Need an nginx samurai
- [x] run all tests using `cli`
- [x] check users `fpm` and `cli` run as, can they collaborate writing on the same volume? Is it just `root`?
- [x] `composer install` command. Customization? Perhaps classmap or not, but save it for later.
- [x] try it on a Mac
- [x] look at the formula and see what we are missing
- [x] `apt-get` end result is cached potentially forever, but we will rebuild the elife base image from scratch skipping cached


.dockerignore
```diff
@@ -0,0 +1,2 @@
+build/*
+.git
```
giorgiosironi: Perhaps there is more we can ignore, but it's purely about performance rather than correctness (unless we are including some secret credentials in the image).

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM php:7.0.26-fpm-jessie
```
giorgiosironi: This is an official (as in official from Docker, not official from the PHP core team) image:
- `7.0` because it's the closest version to what we use now
- `fpm` because it's the web-facing version with `php-fpm` (equivalent to `uwsgi` for the Python folks) and not `cli`. I will have to create a separate `cli` image that instead can be used for the long-running process looking at the SQS queue, and for running all tests (that test also the long-running process and do not work in `fpm`)
- `jessie` (Debian oldstable) because there is no `ubuntu` in the official images. The alternative is `alpine` which while being locked down and lean in size, has a different package management system. I tend to be conservative and go for the most similar option.

In any case, we will need a base `elife/php-fpm` (naming to be decided) image so that all PHP projects share the same base.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+RUN apt-get update && apt-get install -y git-core zip
```
giorgiosironi: Dependencies needed to run `composer`. As the comment says, there could be another way (multi-stage builds) that complicates things.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+# replace with multi-stage build using the composer official image?
+RUN curl https://getcomposer.org/download/1.5.2/composer.phar -o /usr/local/bin/composer && chmod +x /usr/local/bin/composer 
+
+WORKDIR /srv/annotations
+COPY composer.json composer.lock /srv/annotations/
+RUN composer install
```
giorgiosironi: standard practice to resolve dependencies before copying the code, so that all that can be cached when rebuilding the image after a code change

build_image.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+docker build -t annotations .
```
giorgiosironi: creates a local image, necessary to execute `docker run`

config/container.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+return ['containerized' => true];
```
giorgiosironi: This is temporary and should be removed, but was necessary to get logs up and running. 

Standard practice would be to log to `stderr` rather than files, and have the container runtime or the orchestrator direct the logs somewhere. However this [works poorly with php-fpm](https://buff.ly/2ANZTlI) as the logs are cluttered in the best syslog tradition and can't even be parsed. So we'll need to find another way to deal with logs, for example to write them locally as before and have a sidecar container upload them.

For the development environment, `stderr` should be enough however, so I should transform this into a agnostic configuration like `'logging.handlers' => 'stderr'`

config/nginx-default.conf
```diff
@@ -0,0 +1,28 @@
...
+        try_files $uri /app_dev.php$is_args$args;
+    }
+
+    location ~ ^/app_dev\.php(/|$) {
+        fastcgi_split_path_info ^(.+\.php)(/.+)$;
+        include fastcgi_params;
+        fastcgi_param SCRIPT_FILENAME $request_filename;
+        fastcgi_param ENVIRONMENT_NAME dev;
+        fastcgi_intercept_errors on;
+        fastcgi_pass php:9000;
```
giorgiosironi: communicating with `php-fpm` over its standard port rather than via the filesystem

config/nginx-default.conf
```diff
@@ -0,0 +1,28 @@
+server {
+    server_name localhost;
+    listen 80;
+    root /srv/annotations/web;
```
giorgiosironi: this is a bit weird as the nginx container doesn't have this folder, and we don't mount it. But it's passed to php-fpm along with `app_dev.php` to find the entry point of the application

config/nginx-default.conf
```diff
@@ -0,0 +1,28 @@
...
+        fastcgi_intercept_errors on;
+        fastcgi_pass php:9000;
+        internal;
+    }
+
+    # return 404 for all other php files not matching the front controller
+    # this prevents access to other php files you don't want to be accessible.
+    location ~ \.php$ {
+        return 404;
+    }
```
giorgiosironi: have to find a way to reduce the boilerplate of this file, but it's early to say. An orchestrator like Kubernetes would have its own way to create configuration files and mount them in container, but we are just talking about local usage here.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+
+services:
+    web:
```
giorgiosironi: next services to add:

- `goaws` for SQS simulation
- separate `php-cli` for the long-running process

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+        image: nginx:1.13.7
```
giorgiosironi: the version we use is not available, had to upgrade

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+        ports:
+            - "8080:80"
+        volumes:
+            - ./config/nginx-default.conf:/etc/nginx/conf.d/default.conf
+            # can we avoid this by removing try_files?
```
giorgiosironi: good question

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+
+services:
+    web:
+        image: nginx:1.13.7
+        ports:
+            - "8080:80"
+        volumes:
+            - ./config/nginx-default.conf:/etc/nginx/conf.d/default.conf
```
giorgiosironi: configuration is not hardcoded into the image, but mounted from outside. That will be necessary for anything containing credentials, and for anything that changes between environments.

giorgiosironi:
Tagged @lsh-0 for a first look at the mess. Maybe some of these files can be extracted into the formula, but the coupling is very high at the moment so I'm trying to keep everything together until stable and postpone that decision.

lsh-0:
thankyou for pre-empting my grumbling ;)

Dockerfile
lsh-0: can you have a dual php-fpm/cli base?

config/nginx-default.conf
```diff
@@ -0,0 +1,28 @@
...
+
+    access_log /var/log/nginx/access.log combined;
```
lsh-0: should these be going to stdout/err ?

src/Annotations/AppKernel.php
```diff
@@ -54,6 +54,7 @@ public function __construct(string $environment = 'dev')
 
         $this->app = new Application([
             'debug' => $config['debug'] ?? false,
+            'containerized' => $config['containerized'] ?? false,
```
lsh-0: is the `$somevar ?? false` syntax the same as `$somevar || false` or `somevar or False` in Python? I remember php had/has the `somecond? true: false` type syntax.

config/nginx-default.conf
giorgiosironi: Yes, that seems the easiest. They would be merged into a single stream which is weird, but it seems even in Kubernetes the granularity of logs would be per-container.

An update on application logs is that (if permissions and sharing of the filesystem can be figured out) a sidecar container can be used to tail the application log files: https://kubernetes.io/docs/concepts/cluster-administration/logging/#streaming-sidecar-container

src/Annotations/AppKernel.php
giorgiosironi: it's sugar over `isset()` usage http://php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op which allows for any expression to fall back to a default value

giorgiosironi:
The deployment model is still in flux, I'll try to write down here a snapshot of what we know so far:

> The only issue I have is with how we are using Docker. It goes back to that point the other day, about to what purpose are we building these images - for development or testing or deployment? You replied with 'for development first, as always', which didn't make much sense to me.

"development first" as in "environments with the lowest risk first" https://blog.codeship.com/growing-tech-stack-say-no/
It's a natural follow-up however that once you use a container locally, its configuration could diverge from what is being tested in `ci` and run in `end2end` or `prod`. So eventually it can eat more of the pipeline and a container being deployed in the other environments, starting from the testing ones.

> In your comments you mention having multiple Dockerfiles, one for fpm, another for cli. I'm ok with having multiple docker files, but how to organise them so their intent is clear?

Convention I have found so far is not to privilege any of them and name them `Dockerfile.fpm` and `Dockerfile.cli`. Not all project would have that, but when you have both a web pool and long-running processes they are effectively using two different parts of PHP to work (similar for Python, only half of the application needs `uwsgi`).

> They can't all share the Dockerfile file and descendants will have to specify a base image. Does this base file live alongside the project, or with the project's formula? 

Not sure where to put the base `php` images yet, but I was thinking of them in the context of multiple projects. To avoid spreading knowledge and basic steps all over different repositories, two `elife/php-fpm` and `elife/php-cli` images can be created.

> How often are we rebuilding these images? Because the apt-get update at the beginning of the file will only ever be run once, meaning any future security patches will be left out. So, I know I've gone completely offtopic, there are just many blank spaces in my head about the deployment process for these containers.

Right now, there is a single Dockerfile so every new commit on `develop` for this project would trigger a full rebuild. I don't expect to change the base official image very often, but it can be changed to get new versions of PHP with bug fixes for example.

Once a base image is extracted, we'll need a way to tag it. We could publish `elife/php-fpm` with tags such as:
- `2fd4e1c6...` (a sha, not very meaningful but precise)
- `0.1.0`, `0.1.1` (some kind of versioning, but overhead of deciding how to assign new numbers)
- `40`, `42`, `45` (passed Jenkins build numbers, monotonically increasing but unnecessarily tied to the build system)
- `20171205` (date-based versioning, seems the best solution so far)

Then it's a matter of updating a dependency: by default a commit will always build with the same base image, but a dependency-updating build like https://alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-api-raml/ can open a pull request as often as needed. The frequency of these builds goes from weekly to daily depending on the dependency stability and the overhead.

There's another, additional matter that is the caching of the intermediate image resulting from the apt-get. Need to investigate more about that.

giorgiosironi:
I have found https://hub.docker.com/u/elifesciences/ and changed its password to put into it-admin. We could turn that user into an organization? It seems it's still free of charge as long as the repositories are public.

giorgiosironi:
@lsh-0 I found myself looking for solutions to the `apt-get update` problem and my conclusion so far is that when we update the base image, we bring in the security updates. That update should be made while disabling Docker's image cache.

This will be food for an ADR somewhere soon, but in general I'd say:
- never run `apt-get upgrade` inside a Dockerfile (in general)
- always run `apt-get update` and `apt-get install` inside a Dockerfile, if you need additional packages (in general)
- extract common packages into a base image like `elife/php` (possible eLife policy)
- get updates by rebuilding the base `elife/php` with no cache at all and with its `latest` version (possible eLife policy, as a consequence)

Collected more general thoughts on this in https://unix.stackexchange.com/a/408954/95825

giorgiosironi:
Ready for review, going to leave this here for a while.

.dockerignore
```diff
@@ -0,0 +1,4 @@
+build/*
```
thewilkybarkid: I _think_ this wouldn't ignore anything in sub-directories (so better not to have `*`).

.dockerignore
```diff
@@ -0,0 +1,4 @@
...
+.git
+docker-compose.yml
+var/
```
thewilkybarkid: Missing `config.php` and `vendor/`?

Dockerfile.cli
```diff
@@ -0,0 +1,16 @@
+FROM php:7.0.26-cli-jessie
```
thewilkybarkid: `7.0.26` is very specific.

Dockerfile.cli
```diff
@@ -0,0 +1,16 @@
...
+RUN docker-php-ext-install pcntl
+RUN apt-get update && apt-get install -y git-core zip
+# try apt-get clean
+
+# replace with multi-stage build using the composer official image?
+RUN curl https://getcomposer.org/download/1.5.2/composer.phar -o /usr/local/bin/composer && chmod +x /usr/local/bin/composer
```
thewilkybarkid: `1.5.2` is very specific.

src/Annotations/AppKernel.php
```diff
@@ -48,7 +48,10 @@
     public function __construct(string $environment = 'dev')
     {
         $configFile = __DIR__.'/../../config.php';
-        $config = array_merge(file_exists($configFile) ? require $configFile : [], require __DIR__."/../../config/{$environment}.php");
+        $config = array_merge(
+            require __DIR__."/../../config/{$environment}.php",
+            file_exists($configFile) ? require $configFile : []
```
thewilkybarkid: Is the order change meant?

giorgiosironi: Yes, because otherwise it's not possibly to override an environment's committed file with the provisioned one.

thewilkybarkid: What's the use case? (Can understand `dev` for trying stuff out, but sounds dangerous for any other environment.)

giorgiosironi: (`container.php` `aws` setting would be ignored)

Dockerfile.cli
giorgiosironi: It is as there is no `Dockerfile.lock`. `7.0-cli` (which defaults to `jessie`) should maintain the current behavior instead, where a new release of the code is forced to adopt a new patch version of the `php` package.

We are going to end up having a elife php image extracted from here that we can version and update periodically, so it should not be necessary to lock the patch version here.

src/Annotations/AppKernel.php
thewilkybarkid: Problem to me looks to be that `dev` has hard-coded configuration.

Dockerfile.cli
thewilkybarkid: :+1: to an eLife image.

Dockerfile.cli
giorgiosironi: Should we port https://github.com/elifesciences/builder-base-formula/blob/master/elife/composer.sls#L53-L71 here or is there an updated .phar URL?

thewilkybarkid: That's the recommended way to install. AFAIK there's no `1.x` URI (maybe we could request one?).

src/Annotations/AppKernel.php
giorgiosironi: It seems to contain the defaults for the Vagrant VM, which are also the defaults for running the project locally with the help of `goaws`. In fact, `end2end`|`continuumtest`|`prod` are all empty as everything is provided by the provisioned `config.php`.

.dockerignore
thewilkybarkid: (I know the syntax is different to `.gitignore`; not actually tried so could be wrong.)

Dockerfile.cli
giorgiosironi: Needs to be extracted from here after this container has been integrated into multiple environments, as we are likely to discover more problems.

.dockerignore
```diff
@@ -0,0 +1,6 @@
+build/**/*
```
lsh-0: wouldn't just `build/` do the same job here?

giorgiosironi: probably yes, still confused between the .gitignore syntax and this one

lsh-0: oh wait - `.dockerignore` - sorry, I mistook it as well. could be correct for all I know :P

giorgiosironi: in the end it was correct to use `build/` as the other form is only necessary to exlude, say, all `*.php` files from any of the subdirectory (`build/**/*.php`) but not other files
in `.gitignore` this would be `/build/

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+    logger:
+        image: busybox:1.27.2
+        volumes:
+            - logs:/logs
+        # be careful, this will only tail alredy existing files
+        command: tail -f /logs/all.json
+        depends_on:
+            - cli
+    # TODO: api-dummy
+    # TODO: hypothesis-dummy
```
giorgiosironi: postponing those two

giorgiosironi:
Will redo the setup on a Mac PhpStorm to cross-check it goes well. Running the application and tests worked well in Linux, while xdebug needs a separate PR.

giorgiosironi:
Deploy and running PHPUnit tests on a Mac PhpStorm went well. No more changes planned here.

giorgiosironi:
Anything else needed here? Would like to follow up by building the container in `ci`, try to use it to run tests etc. there, but not in this overly long PR.

giorgiosironi:
It looks like the formula provides a `config.php` that even in `dev` (Vagrant VM only) already contains everything that's needed. Therefore only `container.php` needs to contain some actual values.

giorgiosironi:
About the scripts:
- `dockerized-*` are mostly demonstrative now, to show how `docker-compose` can be used. They can be dropped?
- `scripts/install-composer.sh` should go into the base `elife/php` image(s) when extracted, I don't expect it to stay around in a project.
- `build_images.sh` is what is really necessary since it specifies what to build. This could become a standard like `project_tests.sh`

thewilkybarkid:
`dockerized-*` should probably go. Good to see how to use `docker-compose`, but should be documentation rather than `dockerized-` all-the-things.

`build_images.sh` is fine; `scripts/install-composer.sh` is fine for now (but agree that Composer should be part of the image).

src/Annotations/AppKernel.php
giorgiosironi: It doesn't make much of a difference now because the `config/$env.php` files are almost empty. I'll revert it and later I'll remove everything remaining in them until this only loads `config.php`

PR #33, opened 2017-12-12
Extend credentials to allow for jwt signing

nlisgo:
In order to access annotations for a user we must authenticate as that user. In order to do that we need to request an access token. There are 2 types of credential `user_management` and `jwt_signing`.

If this is accepted we must make some changes to the formula to add the `jwt_signing` credentials.


composer.lock
```diff
@@ -495,7 +495,7 @@
                 "MIT"
             ],
             "description": "eLife Sciences API client",
-            "time": "2017-10-23T09:04:22+00:00"
+            "time": "2017-10-23 09:04:22"
```
thewilkybarkid: Are you using an out-of-date version of Composer?

src/HypothesisClient/ApiClient/ApiClient.php
```diff
@@ -13,7 +13,7 @@
 {
     private $httpClient;
     private $headers;
-    private $credentials;
+    private $credentials = null;
```
thewilkybarkid: That's redundant.

src/HypothesisClient/ApiClient/UsersClient.php
```diff
@@ -52,11 +53,54 @@ public function updateUser(
...
+                    'assertion' => $this->getCredentials()->getJWT($id),
+                ],
+            ])
+        );
+    }
+
+    public function getUserAnnotations(
+        array $headers,
+        string $user,
+        $token = null,
```
thewilkybarkid: `string`?

src/HypothesisClient/ApiClient/UsersClient.php
```diff
@@ -52,11 +53,54 @@ public function updateUser(
...
+        int $page = 1,
+        int $perPage = 20,
+        bool $descendingOrder = true,
+        $group = '__world__',
+        $restricted = false
```
thewilkybarkid: `bool`

src/HypothesisClient/ApiClient/UsersClient.php
```diff
@@ -52,11 +53,54 @@ public function updateUser(
...
+    }
+
+    public function getUserAnnotations(
+        array $headers,
+        string $user,
+        $token = null,
+        int $page = 1,
+        int $perPage = 20,
+        bool $descendingOrder = true,
+        $group = '__world__',
```
thewilkybarkid: `string`

src/HypothesisClient/ApiClient/UsersClient.php
```diff
@@ -52,11 +53,54 @@ public function updateUser(
...
+                'path' => 'search',
+                'query' => build_query([
+                    'user' => $user,
+                    'group' => $group,
+                    'offset' => ($page - 1) * $perPage,
+                    'limit' => $perPage,
+                    'order' => $descendingOrder ? 'desc' : 'asc',
+                ]),
+            ]),
+            (($restricted && $token) ? ['Authorization' => 'Bearer '.$token] : []) + $headers
```
thewilkybarkid: Is `$restricted` needed?

tests/HypothesisClient/Credentials/UserManagementCredentialTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @covers \eLife\HypothesisClient\Credentials\UserManagementCredential
+ */
+class UserManagementCredentialTest extends PHPUnit_Framework_TestCase
+{
+    /**
+     * @test
+     */
+    public function it_has_getters()
```
thewilkybarkid: That's rather meaningless.

src/HypothesisClient/ApiClient/ApiClient.php
```diff
@@ -27,14 +27,21 @@ public function __construct(HttpClient $httpClient, $credentials = null, array $
     private function setCredentials(Credentials $credentials)
```
thewilkybarkid: Now redundant?

tests/HypothesisClient/Credentials/UserManagementCredentialTest.php
nlisgo: I was just moving an existing test, I will review.

src/HypothesisClient/ApiClient/UsersClient.php
nlisgo: Thinking about this again and I don't like it, we shouldn't be requesting a token if we don't want to see restricted annotations.

src/HypothesisClient/Credentials/Credentials.php
```diff
@@ -2,56 +2,61 @@
...
 {
-    private $clientId;
-    private $secret;
+    private $userManagement;
+    private $jwtSigning;
     private $authority;
+    private $group;
 
-    public function __construct(string $clientId, string $secret, string $authority)
+    public function __construct(UserManagementCredential $userManagement, JWTSigningCredential $jwtSigning, string $authority, string $group)
```
thewilkybarkid: This class seems to be doing multiple things: the two different sets of credentials don't need to be tied together.

src/HypothesisClient/ApiClient/UsersClient.php
```diff
@@ -52,11 +53,54 @@ public function updateUser(
...
+            json_encode([
+                'form_params' => [
+                    'grant_type' => 'urn:ietf:params:oauth:grant-type:jwt-bearer',
+                    'assertion' => $this->getCredentials()->getJWT($id),
+                ],
+            ])
+        );
+    }
+
+    public function getUserAnnotations(
```
thewilkybarkid: Could be in a separate client? Doesn't need to be in the same one as user management?

config.php.example
```diff
@@ -4,9 +4,17 @@ return [
...
         'api_url' => 'https://hypothes.is/api/',
-        'client_id' => '',
-        'secret_key' => '',
+        'user_management' => [
+            'client_id' => '',
+            'secret_key' => '',
+        ],
+        'jwt_signing' => [
+            'client_id' => '',
+            'secret_key' => '',
```
thewilkybarkid: Normally called a `client_secret`.

composer.lock
nlisgo: yep. sorted

src/HypothesisClient/ApiClient/UsersClient.php
thewilkybarkid: These methods should probably match the API end points, so this should be a search method with an optional `user` parameter (and the client itself is configured with a token or not).

src/HypothesisClient/ApiClient/TokenClient.php
```diff
@@ -0,0 +1,49 @@
...
+
+final class TokenClient
+{
+    use ApiClient;
+
+    private $credentials;
+
+    public function __construct(HttpClient $httpClient, JWTSigningCredentials $credentials, array $headers = [])
+    {
+        $this->httpClient = new UserAgentPrependingHttpClient($httpClient, 'HypothesisClient');
```
giorgiosironi: I think we usually prepend the name of the project in the `User-Agent` too 
e.g. `eLifeJournal/dev-master@7b88378 eLifeApiSdk/dev-master@af6b160 eLifeApiClient/dev-master@dba0869 GuzzleHttp/6.2.1 curl/7.35.0 PHP/7.0.23-1+ubuntu14.04.1+deb.sury.org+1`
which here would only need `eLifeAnnotations/dev-master` as the second parameter (let's ignore the commit for now)

src/HypothesisClient/Credentials/JWTSigningCredentials.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function getExpire() : int
+    {
+        return $this->expire;
+    }
+
+    public function getJWT(string $username) : string
+    {
+        $now = $_SERVER['REQUEST_TIME'];
```
giorgiosironi: `$_SERVER` can't be used from the cli (hence from tests) without overriding global variables; inject a `integer` timestamp in the constructor

giorgiosironi: (as this is not used in the long-running process)

src/HypothesisClient/ApiClient/TokenClient.php
nlisgo: Perhaps this should be set in configuration with `HypothesisClient` as the default?

giorgiosironi: Should not be able to change by environment so I think not a good candidate for configuration. Will need revising anyway when/if we extract `eLife\HypothesisClient` as a library from here

src/HypothesisClient/Credentials/JWTSigningCredentials.php
```diff
@@ -0,0 +1,61 @@
...
+namespace eLife\HypothesisClient\Credentials;
+
+use Firebase\JWT\JWT;
+
+class JWTSigningCredentials extends Credentials
+{
+    private $expire;
+    private $startTime;
+
+    public function __construct(string $clientId, string $clientSecret, string $authority, int $expire = 600, int $startTime = null)
```
giorgiosironi: was looking for where `JWTSigningCredentials` is instantiated in the Kernel, but found only the tests
would make `startTime` mandatory rather than default, to force creation code to pass it and express the dependency and non-determinism (moving it up in the list then as `$expire` seems to have a safe default)

tests/HypothesisClient/Credentials/JWTSigningCredentialsTest.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @test
+     */
+    public function it_may_have_a_start_time()
+    {
+        $credentials = new JWTSigningCredentials('foo', 'baz', 'authority');
+        $this->assertGreaterThanOrEqual(time(), $credentials->getStartTime());
+        $start_time = time();
```
giorgiosironi: no need for this to be dynamic throughout tests, for better reproducibility it can be a constant like `1500000000`

src/HypothesisClient/Credentials/JWTSigningCredentials.php
nlisgo: I will be introducing it to the Kernel later. These are necessary refactors to pave the way for the annotations API.

nlisgo: I intend for `JWTSigningCredntials` to be instantiated in `Kernel` and so setting a `startTime` as a constructor argument is not going to work when we need to make token requests in the long running process. We really need the start time to be reset every when we generate the JWT.

giorgiosironi: If a different time value is required each time, a common pattern a is to inject a `Clock` object instead of the integer. A lightweight version of that is to inject a function returning an integer:
`__construct(..., callable $time)`
Then, the class can be instantiated like this:
```
new JWTSigningCredentials(..., 'time'); // Kernel
new JWTSigningCredentials(..., function() { return '1500000000'; }); // tests
```

nlisgo: I have added a `Clock` object.

src/HypothesisClient/Clock/Clock.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace eLife\HypothesisClient\Clock;
+
+class Clock
+{
+    public function time() : int
```
giorgiosironi: the method signature is as expected :+1:

src/HypothesisClient/Credentials/JWTSigningCredentials.php
```diff
@@ -0,0 +1,45 @@
...
+    private $expire;
+
+    public function __construct(string $clientId, string $clientSecret, string $authority, Clock $clock, int $expire = 600)
+    {
+        parent::__construct($clientId, $clientSecret, $authority);
+        $this->clock = $clock;
+        $this->expire = $expire;
+    }
+
+    public function getExpire() : int
```
giorgiosironi: Without seeing the intended usage I can't be sure of this, but `getExpire()` and `getStartTime()` seem two getters that are not used outside of the tests for this class. If the goal is to only call `getJWT` from client code, these two can be removed (and hence their tests too). The advantage is less code to read and maintain, fewer tests, same functionality.

giorgiosironi: (the inherited `getAuthority()` _is_ stricly needed instead)

tests/HypothesisClient/Clock/ClockTest.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+class ClockTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @test
+     */
+    public function it_can_return_the_current_time()
+    {
+        $clock = new Clock();
+        $this->assertEquals(time(), $clock->time());
```
giorgiosironi: This test mirrors the implementation (http://jasonrudolph.com/blog/2008/07/30/testing-anti-patterns-the-ugly-mirror/)
Since the result is non-deterministic by design, I would relax the expectation to be `$this->assertGreaterThan(0, $clock->time());`

tests/HypothesisClient/Credentials/JWTSigningCredentialsTest.php
```diff
@@ -0,0 +1,56 @@
...
+        $credentials = new JWTSigningCredentials('foo', 'baz', 'authority', new Clock());
+        $this->assertGreaterThan(0, $credentials->getExpire());
+        $credentials = new JWTSigningCredentials('foo', 'baz', 'authority', new Clock(), 100);
+        $this->assertEquals(100, $credentials->getExpire());
+    }
+
+    /**
+     * @test
+     */
+    public function it_can_generate_a_jwt_token()
```
giorgiosironi: this is the only test that this class needs imho: one public method with no conditionals or loops, one functionality to cover

tests/HypothesisClient/Credentials/JWTSigningCredentialsTest.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $now = 1500000000;
+        $clock = $this->createMock(Clock::class);
+        $clock
+            ->expects($this->once())
+            ->method('time')
+            ->willReturn($now);
+        $credentials = new JWTSigningCredentials('foo', 'baz', 'authority', $clock, 300);
+        $payload = [
```
giorgiosironi: it has a mirror problem however, so I'll make you a PR to see if I can raise its abstraction level not to require these

thewilkybarkid:
A lot of this doesn't actually seem to be used?

giorgiosironi:
See https://github.com/elifesciences/annotations/pull/33#discussion_r156680472

thewilkybarkid:
Hmm, more than just a refactor. Something for next time.

nlisgo:
@thewilkybarkid I would agree that this pull request could have been even smaller.

We have an `AnnotationsClient` already in the codebase which isn't being used yet (it shouldn't be there perhaps), I want to use it but need to refactor so that I can optionally request restricted annotations.

I could have equally prepared a larger PR which ties all of the client code together but I wanted to invite feedback earlier. I welcome your input on how I could approach it better.

giorgiosironi:
Hypothetical slices:
1. refactor existing credentials in `user_management`, merge and deploy
2. add `jwt_signing` credentials

PR #37, opened 2017-12-19
ELPP-3343 Use a container to run ci

giorgiosironi:
Creates an additional `annotations_ci` container to run proofreader, PHPUnit and smoke tests.

Uses exclusively `docker-compose` to interact with the Docker daemon.

Still runs the old VM-based `Project tests` build after the container build, to show both are working now.

Limitations in scope:

- do not introduce any new OS user (stick to `root` and `www-data`)
- do not introduce any eLife-wide base image (too early for that to be really reusable)
- do not introduce any separate image for multi-stage builds (e.g. to install Composer or smoke tests)
- do not push the resulting images anywhere
  
Workarounds:

- Dockerfiles `COPY` instruction
- Docker's strict volume permissions to take out XML the build result
- PhpStorm limitations (no support for multiple `docker-compose.yml` files to be merged together)

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -9,6 +9,6 @@ RUN mkdir -p build && chown www-data:www-data build
 USER www-data
 RUN composer global --no-interaction require elife/proofreader-php=dev-master
 ENV PATH="/var/www/.composer/vendor/bin:${PATH}"
-COPY project_tests.sh smoke_tests.sh /srv/annotations/
+COPY .php_cs project_tests.sh smoke_tests.sh /srv/annotations/
 
 CMD ["/bin/bash", "project_tests.sh"]
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,4 +1,7 @@
 FROM annotations_cli
+USER root
+RUN mkdir -p build && chown www-data:www-data build
+USER www-data
 RUN composer global --no-interaction require elife/proofreader-php=dev-master
 ENV PATH="/var/www/.composer/vendor/bin:${PATH}"
 CMD ["/bin/bash", "project_tests.sh"]
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,0 +1,4 @@
+FROM annotations_cli
+RUN composer global --no-interaction require elife/proofreader-php=dev-master
+ENV PATH="/var/www/.composer/vendor/bin:${PATH}"
+CMD ["/bin/bash", "project_tests.sh"]
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,7 +1,14 @@
 FROM annotations_cli
 USER root
+
+# extract into another image
+RUN git clone https://github.com/asm89/smoke.sh /opt/smoke.sh
+
 RUN mkdir -p build && chown www-data:www-data build
+
 USER www-data
 RUN composer global --no-interaction require elife/proofreader-php=dev-master
 ENV PATH="/var/www/.composer/vendor/bin:${PATH}"
+COPY project_tests.sh smoke_tests.sh /srv/annotations/
+
 CMD ["/bin/bash", "project_tests.sh"]
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -8,12 +8,18 @@ RUN /root/scripts/install-composer.sh
 
 WORKDIR /srv/annotations
 COPY composer.json composer.lock /srv/annotations/
-# could probably use --no-dev and other optimizations
+# could probably use --no-dev and other optimizations,
 # if we specialize annotations_ci instead
 RUN composer install
 
 # TODO: remove phpunit.xml.dist from here?
-COPY bin config phpunit.xml.dist src tests web /srv/annotations/
+# yes this is how you copy directories
+COPY bin/ /srv/annotations/bin/
+COPY config/ /srv/annotations/config/
+COPY phpunit.xml.dist /srv/annotations/
+COPY src/ /srv/annotations/src/
+COPY tests/ /srv/annotations/tests/
+COPY web/ /srv/annotations/web/
 
 RUN mkdir -p var/logs && chown www-data:www-data var/logs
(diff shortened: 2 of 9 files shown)


Dockerfile.ci
```diff
@@ -0,0 +1,14 @@
+FROM annotations_cli
+USER root
```
giorgiosironi: we don't *have* to use `root`, but I'd like to have a different user than `www-data` so that files are not modifiable from the PHP processes

Dockerfile.ci
```diff
@@ -0,0 +1,14 @@
...
+
+# extract into another image
```
giorgiosironi: possibly with multistage builds

Dockerfile.ci
giorgiosironi: possible solution: just use `www-data` now that we target `dev` and `ci` and worry about it later when `end2end`|`prod` images are produced

Dockerfile.cli
```diff
@@ -8,9 +8,20 @@ RUN /root/scripts/install-composer.sh
 
 WORKDIR /srv/annotations
 COPY composer.json composer.lock /srv/annotations/
+# could probably use --no-dev and other optimizations,
+# if we specialize annotations_ci instead
 RUN composer install
 
-COPY . /srv/annotations
+# TODO: remove phpunit.xml.dist from here?
+# yes this is how you copy directories
```
giorgiosironi: alternative is to `COPY . /srv/annotations` and relying on .dockerignore, but it cannot exclude files that only need to be sent to one image of the 3 like `smoke_tests.sh`

Dockerfile.fpm
```diff
@@ -1,5 +1,5 @@
 FROM php:7.0-fpm-jessie
-
+RUN docker-php-ext-install pcntl
```
giorgiosironi: loaded by the bootstrap of the application, so needed also here

Jenkinsfile
```diff
@@ -1,9 +1,21 @@
...
+                sh 'docker-compose build'
+                sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh'
+                step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
+                sh 'docker-compose -f docker-compose.ci.yml run ci ./smoke_tests.sh web'
+            }
+        },
+        'elife-libraries--ci'
+    )
 
     stage 'Project tests', {
```
giorgiosironi: the `Project tests` stage will be removed in favor of `Container image` (which will assume that name)

Dockerfile.ci
thewilkybarkid: Sounds like something to park. The future base eLife image would presumably come with an `elife` user.

Jenkinsfile
```diff
@@ -1,9 +1,21 @@
...
+    elifeOnNode(
+        {
+            stage 'Checkout', {
+                checkout scm
+                commit = elifeGitRevision()
+            }
+
+            stage 'Container image', {
+                sh 'docker-compose build'
+                sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh'
```
giorgiosironi: 777 is necessary because the container is then able to write build artifacts like XML test results in `build/`

Jenkinsfile
```diff
@@ -1,9 +1,21 @@
...
+                step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
+                sh 'docker-compose -f docker-compose.ci.yml run ci ./smoke_tests.sh web'
```
giorgiosironi: `web` is the hostname of the nginx container

build_images.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
```
giorgiosironi: superseded by `docker-compose build`

docker-compose.ci.yml
```diff
@@ -0,0 +1,71 @@
+version: '3'
+
+
+services:
```
giorgiosironi: lots of duplication with `dev`, I would extract the common stuff into a `docker-compose.base.yml`, but PhpStorm doesn't support multiple docker-compose.yml files and I wanted to keep that route open

docker-compose.yml
```diff
@@ -15,16 +15,20 @@ services:
...
             dockerfile: Dockerfile.fpm
+        image: annotations_fpm
         volumes:
             - ./:/srv/annotations
             - ./config/container.php:/srv/annotations/config.php
             - logs:/srv/annotations/var/logs
         env_file:
             - dev.env
+        depends_on:
+            - goaws
```
giorgiosironi: loads the queue name in the bootstrap, so needed

docker-compose.yml
```diff
@@ -15,16 +15,20 @@ services:
         build: 
             context: .
             dockerfile: Dockerfile.fpm
+        image: annotations_fpm
```
giorgiosironi: image names are used by `docker-compose build`

Dockerfile.cli
thewilkybarkid: `rm`?

smoke_tests.sh
```diff
@@ -4,9 +4,13 @@ set -e
 
 bin/console --version
 
+host="${1:-$(hostname)}"
```
giorgiosironi: run from the `ci` container, targets the `web` container passed as an argument

Dockerfile.cli
thewilkybarkid: What happens if you copy an ignored file?

giorgiosironi: the build phase is executed by the Docker daemon, which is not sent the ignored files, meaning the file is not there to be copied

Dockerfile.ci
giorgiosironi: Checked out both the `root` road and the `www-data` road and it seems the first is the simplest as it's the default user of the Dockerfiles (while the third, `elife`, will be introduced later)

PR #38, opened 2018-01-09
Remove annotations--ci EC2 instance

giorgiosironi:
One less VM to maintain, and by consolidating on fewer instances we get faster builds as it may already be in a running state rather than shutdown.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+COMPOSE_PROJECT_NAME=annotations
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,6 +9,9 @@ elifePipeline {
 
             stage 'Container image', {
                 sh 'docker-compose -f docker-compose.ci.yml build'
+            }
+
+            stage 'Project tests', {
                 sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh'
                 step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
                 sh 'docker-compose -f docker-compose.ci.yml run ci ./smoke_tests.sh web'
@@ -16,13 +19,6 @@ elifePipeline {
         },
         'elife-libraries--ci'
     )
-
-    stage 'Project tests', {
-        lock('annotations--ci') {
-            builderDeployRevision 'annotations--ci', commit
-            builderProjectTests 'annotations--ci', '/srv/annotations', ['/srv/annotations/build/phpunit.xml']
-        }
-    }
 
     elifeMainlineOnly {
         stage 'End2end tests', {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,9 +12,9 @@ elifePipeline {
             }
 
             stage 'Project tests', {
-                sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh'
+                sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run --rm ci ./project_tests.sh'
                 step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
-                sh 'docker-compose -f docker-compose.ci.yml run ci ./smoke_tests.sh web'
+                sh 'docker-compose -f docker-compose.ci.yml run --rm ci ./smoke_tests.sh web'
             }
         },
         'elife-libraries--ci'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ elifePipeline {
 
     elifeOnNode(
         {
-            stage 'Container image', {
+            stage 'Build images', {
                 checkout scm
                 sh 'docker-compose -f docker-compose.ci.yml build'
             }
@@ -19,6 +19,15 @@ elifePipeline {
                     sh 'docker-compose -f docker-compose.ci.yml run --rm ci ./smoke_tests.sh web'
                 } finally {
                     sh 'docker-compose -f docker-compose.ci.yml stop'
+                }
+            }
+
+            elifeMainlineOnly {
+                stage 'Push images', {
+                    sh "docker tag annotations_cli elifesciences/annotations_cli:latest && docker push elifesciences/annotations_cli:latest"
+                    sh "docker tag annotations_cli elifesciences/annotations_cli:${commit} && docker push elifesciences/annotations_cli:${commit}"
+                    sh "docker tag annotations_fpm elifesciences/annotations_fpm:latest && docker push elifesciences/annotations_fpm:latest"
+                    sh "docker tag annotations_fpm elifesciences/annotations_fpm:${commit} && docker push elifesciences/annotations_fpm:${commit}"
                 }
             }
         },
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,13 +1,14 @@
 elifePipeline {
     def commit
+    stage 'Checkout', {
+        checkout scm
+        commit = elifeGitRevision()
+    }
+
     elifeOnNode(
         {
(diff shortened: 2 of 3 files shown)


giorgiosironi:
Interesting failure:
```
ERROR: for onspr3862bycpvmeteo33pgcb54_goaws_1  Cannot start service goaws: driver failed programming external connectivity on endpoint onspr3862bycpvmeteo33pgcb54_goaws_1 (e1e8ab66927b001b34fcf1b728b829048b95742d27e18990dc1edbc591834dce): Bind for 0.0.0.0:4100 failed: port is already allocated
```
Something (probably another `goaws`) is using port 4100 on the libraries node.

giorgiosironi:
Containers from a previous build are still up and running:
```
elife@ci--libraries:~$ sudo docker container ls
CONTAINER ID        IMAGE                            COMMAND                  CREATED             STATUS              PORTS                    NAMES
dd34468f6ae1        nginx:1.13.7                     "nginx -g 'daemon ..."   9 minutes ago       Up 9 minutes        0.0.0.0:8080->80/tcp     testannotations_web_1
89a8f72a6dde        annotations_fpm                  "docker-php-entryp..."   9 minutes ago       Up 9 minutes        9000/tcp                 testannotations_fpm_1
68da6df67953        annotations_cli                  "docker-php-entryp..."   9 minutes ago       Up 9 minutes                                 testannotations_cli_1
73c44ae36ac3        elifealfreduser/goaws:20171024   "/goaws_linux_amd64"     9 minutes ago       Up 9 minutes        0.0.0.0:4100->4100/tcp   testannotations_goaws_1
```

Jenkinsfile
```diff
@@ -1,29 +1,30 @@
...
+    stage 'Checkout', {
+        checkout scm
+        commit = elifeGitRevision()
+    }
+
     elifeOnNode(
         {
-            stage 'Checkout', {
+            stage 'Container image', {
                 checkout scm
```
giorgiosironi: Jenkins performs everything inside `elifeOnNode()` on a separate node, that's why we need an additional `checkout scm` that performs the git clone on this node too. In the future we could do everything on this `elife-libraries--...` node, but for now the rest of the build after `ci` still needs to be executed on the Jenkins main node.

Jenkinsfile
```diff
@@ -1,29 +1,30 @@
...
-                sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh'
-                step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
-                sh 'docker-compose -f docker-compose.ci.yml run ci ./smoke_tests.sh web'
+            stage 'Project tests', {
+                try {
+                    sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run --rm ci ./project_tests.sh'
+                    step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
+                    sh 'docker-compose -f docker-compose.ci.yml run --rm ci ./smoke_tests.sh web'
+                } finally {
+                    sh 'docker-compose -f docker-compose.ci.yml stop'
```
giorgiosironi: Otherwise the containers stay up, occupying CPU, memory, ports and preventing further builds from running.

Jenkinsfile
```diff
@@ -1,29 +1,30 @@
...
+                }
             }
         },
         'elife-libraries--ci'
     )
 
-    stage 'Project tests', {
```
giorgiosironi: The only stage that depended on `annotations--ci`.

Jenkinsfile
```diff
@@ -1,29 +1,39 @@
...
+                    sh 'chmod 777 build/ && docker-compose -f docker-compose.ci.yml run --rm ci ./project_tests.sh'
+                    step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
+                    sh 'docker-compose -f docker-compose.ci.yml run --rm ci ./smoke_tests.sh web'
+                } finally {
+                    sh 'docker-compose -f docker-compose.ci.yml stop'
+                }
+            }
+
+            elifeMainlineOnly {
+                stage 'Push images', {
```
giorgiosironi: they end up at:
- https://hub.docker.com/r/elifesciences/annotations_cli/tags/
- https://hub.docker.com/r/elifesciences/annotations_fpm/tags/

Jenkinsfile
```diff
@@ -1,29 +1,39 @@
...
+                    sh "docker tag annotations_cli elifesciences/annotations_cli:latest && docker push elifesciences/annotations_cli:latest"
+                    sh "docker tag annotations_cli elifesciences/annotations_cli:${commit} && docker push elifesciences/annotations_cli:${commit}"
+                    sh "docker tag annotations_fpm elifesciences/annotations_fpm:latest && docker push elifesciences/annotations_fpm:latest"
+                    sh "docker tag annotations_fpm elifesciences/annotations_fpm:${commit} && docker push elifesciences/annotations_fpm:${commit}"
```
giorgiosironi: all the `docker` and `docker-compose` commands are getting complex, but if they are extracted now they will be hidden in https://github.com/elifesciences/elife-jenkins-workflow-libs so better to extract them once stable and reusable

.env
```diff
@@ -0,0 +1 @@
+COMPOSE_PROJECT_NAME=annotations
```
giorgiosironi: this is the prefix for local container images (and the resulting containers):
```
annotations_ci                        latest                                     83d6f50205b5        24 hours ago        493MB
annotations_cli                       latest                                     e150cfc63934        3 weeks ago         447MB
annotations_fpm                       latest                                     7b39318a9a47        3 weeks ago         459MB
```

Jenkinsfile
```diff
@@ -1,29 +1,39 @@
 elifePipeline {
     def commit
+    stage 'Checkout', {
```
giorgiosironi: Everything inside `elifeNode()` is performed on a separate node with Docker installed and a large disk to store images. We do `checkout scm` (which is a `git clone`) on both the main node and this separate `elife-libraries--*` one because both need access to either the repository files or Git metadata such as the commit sha.

Jenkinsfile
giorgiosironi: The idea is downstream servers will be able to pull the images without building them from source, shaving off minutes from every pipeline run. At the moment we are still targeting EC2 instances running only one application, but once the image-based deployment works we can do more.
Of course we only want to do this for passing builds of `develop` (`elifeMainlineOnly`). See also https://github.com/elifesciences/elife-libraries-formula/pull/5 for the credentials to do this.

docker-compose.ci.yml
```diff
@@ -17,7 +17,7 @@ services:
         build: 
             context: .
             dockerfile: Dockerfile.fpm
-        image: annotations_fpm
+        image: elifesciences/annotations_fpm
```
giorgiosironi: not strictly necessary but matches the remote Docker Hub repository

Jenkinsfile
giorgiosironi: Hence should solve the failure of https://alfred.elifesciences.org/job/test-annotations/32

lsh-0:
so to summarize, and apologies if I get it wrong, annotations has now been migrated to use a docker container to run it's `--ci` tests instead of a separate ec2 instance? and is the annotations--prod instance going to be containerised (eventually)?

giorgiosironi:
> annotations has now been migrated to use a docker container to run it's --ci tests instead of a separate ec2 instance

Yes, but the `ci` instance has not been destroyed yet.

> is the annotations--prod instance going to be containerised (eventually)?

Yes, as one of the OKRs for Q1 2018, but since we don't have an orchestrator yet this would first take the form of an EC2 instance running the pair of containers.

PR #40, opened 2018-01-10
ELPP-3343 Use elife-base-images

giorgiosironi:
Without pinning to a tag at the moment, just the latest.

Depends on https://github.com/elifesciences/elife-base-images/pull/1 but I built and pushed those manually.


giorgiosironi:
#42 depends on this, so I'll merge this change so that these commits can be deleted from there.

PR #41, opened 2018-01-11
Mock requests to Hypothes.is

giorgiosironi:
- `hypothesis.guzzle.handler` was not mocked
- `ValidatingStorageAdapter` can't be used as-is because the Hypothes.is API does not follow its expected format
- Adding response body in error message
- Adding missing DOM extension

### DIFF
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -1,6 +1,7 @@
 FROM php:7.0-cli-jessie
+RUN apt-get update && apt-get install -y git-core zip libxml2-dev
 RUN docker-php-ext-install pcntl
-RUN apt-get update && apt-get install -y git-core zip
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Had to merge the `develop` branch to run the build.

Dockerfile.cli
```diff
@@ -1,6 +1,7 @@
 FROM php:7.0-cli-jessie
+RUN apt-get update && apt-get install -y git-core zip libxml2-dev
 RUN docker-php-ext-install pcntl
-RUN apt-get update && apt-get install -y git-core zip
+RUN docker-php-ext-install dom
```
giorgiosironi: I'm not sure if this is used only in tests (PHPUnit requires it?) or whether it should be put everywhere

PR #42, opened 2018-01-16
Annotations api

nlisgo:


src/Annotations/Serializer/CommonMark/functions_include.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+// Don't redefine the functions if included multiple times.
+if (!function_exists('eLife\Annotations\Serializer\CommonMark\clean_paragraph')) {
+    require __DIR__.'/functions.php';
```
giorgiosironi: use `require_once`?

giorgiosironi: What is the problem in referring directly to `functions.php` in the autoload definition?

src/Annotations/AppKernel.php
```diff
@@ -269,6 +350,70 @@ public function __construct(string $environment = 'dev')
...
+                'application/vnd.elife.annotation-list+json; version=1'
+            ));
+
+        $this->app->after(function (Request $request, Response $response, Application $app) {
+            if ($response->isCacheable()) {
+                $response->headers->set('ETag', md5($response->getContent()));
+                $response->isNotModified($request);
+            }
+
+            if (!$this->app['mock'] && $this->app['debug']) {
```
giorgiosironi: If I get this right, when `mock` is enabled during test we don't validate the application responses. Perhaps a TODO comment can be inserted to try to remove `mock` and do the validation in PHPUnit runs?

src/Annotations/Serializer/AnnotationNormalizer.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * @param Annotation $object
+     */
+    public function normalize($object, $format = null, array $context = []) : array
+    {
+        $content = $object->getText() ? $this->processText($object->getText()) : null;
+
+        $data = array_filter([
```
giorgiosironi: `array_filter` without a callback removes values that are `false`. But which of these values can be `false`? If it's only `parents` it can be inserted separately into the already created `$data`

src/Annotations/Serializer/AnnotationNormalizer.php
```diff
@@ -0,0 +1,157 @@
...
+            ],
+        ]) + ['parents' => []];
+        if ($data['created'] === $data['updated']) {
+            unset($data['updated']);
+        }
+        if ($object->getTarget()->getSelector()) {
+            $data['highlight'] = $object->getTarget()->getSelector()->getTextQuote()->getExact();
+        }
+        if (empty($data['highlight']) && empty($data['content'])) {
+            $this->logger->warning(sprintf('Annotation detected without highligt or content (ID: %s)', $data['id']), ['annotation' => $data]);
```
giorgiosironi: typo in `highligt`

giorgiosironi: can probably be a level higher than `warning`? Seems a bad situation

src/Annotations/Serializer/AnnotationNormalizer.php
```diff
@@ -0,0 +1,157 @@
...
+                        'text' => $rendered,
+                    ];
+                    break;
+                case $block instanceof Element\FencedCode:
+                case $block instanceof Element\IndentedCode:
+                    $data[] = [
+                        'type' => 'code',
+                        'code' => $rendered,
+                    ];
+                    break;
```
giorgiosironi: no `default`? Either a default action (or a `throw SomeException` if it's never meant to get there) would be useful

src/Annotations/Serializer/AnnotationNormalizer.php
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+
+        if (empty($data)) {
+            $data = [
+                [
+                    'type' => 'paragraph',
+                    'text' => self::CANNOT_RENDER_CONTENT_COPY,
```
giorgiosironi: `self::CANNOT_RENDER_CONTENT_COPY` is associated to a log being written in the previous case, perhaps also here it should be?

src/HypothesisClient/Client/Search.php
```diff
@@ -0,0 +1,79 @@
...
+                return $result;
+            })
+            ->then(function (Result $result) {
+                return array_map(function (array $annotation) {
+                    return $this->serializer->denormalize($annotation, Annotation::class);
+                }, $result['rows']);
+            });
+    }
+
+    public function count(
```
giorgiosironi: the two methods can be called in any order, which reflect statelessness. `$this->total` is however a cache that is never invalidated, is it strictly necessary or is it a performance optimization?

tests/HypothesisClient/Clock/SystemClockTest.php
```diff
@@ -0,0 +1,26 @@
...
+     * @test
+     */
+    public function it_can_return_the_current_time()
+    {
+        $clock = new SystemClock();
+        $this->assertInstanceOf(Clock::class, $clock);
+        $time = $clock->time();
+        $this->assertGreaterThan(0, $time);
+        sleep(1);
+        $this->assertNotEquals($time, $clock->time());
```
giorgiosironi: could `assertGreaterThan` also here, but I wouldn't go that far; stop before sleep() for test efficiency, this class is a single line that will already be exercised

tests/HypothesisClient/Clock/FixedClockTest.php
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * @test
+     */
+    public function it_can_return_a_fixed_time()
+    {
+        $clock = new FixedClock();
+        $this->assertInstanceOf(Clock::class, $clock);
+        $time = $clock->time();
+        $this->assertGreaterThan(0, $time);
+        sleep(1);
```
giorgiosironi: to avoid this sleep, pass a fixed value to the constructor of `FixedClock` and `assertEquals(..., $clock->time())` with that value as expected

src/Annotations/Serializer/AnnotationNormalizer.php
```diff
@@ -0,0 +1,157 @@
...
+use eLife\ApiSdk\ApiSdk;
+use eLife\HypothesisClient\Model\Annotation;
+use League\CommonMark\Block\Element;
+use League\CommonMark\DocParser;
+use League\CommonMark\ElementRendererInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
```
giorgiosironi: This class is very important as it marks the boundary between the `eLife\HypothesisClient` subdomain and the `eLife\Annotations` one. It's not to be confused with `AnnotationDenormalizer` and the others `Denormalizer`s.
Therefore it could be given a more unique name, like `eLife\Annotations\Serializer\HypothesisClientAnnotationNormalizer`.

giorgiosironi: There's probably a better name than that.

tests/Annotations/Serializer/AnnotationNormalizerTest.php
```diff
@@ -0,0 +1,689 @@
...
+    private $htmlRenderer;
+    /** @var BufferingLogger */
+    private $logger;
+    /** @var AnnotationNormalizer */
+    private $normalizer;
+
+    /**
+     * @before
+     */
+    protected function setUpNormalizer()
```
giorgiosironi: I agree that this class should be tested in integration with CommonMark

tests/Annotations/Serializer/AnnotationNormalizerTest.php
```diff
@@ -0,0 +1,689 @@
...
+                    'created' => $createdDate,
+                    'document' => [
+                        'title' => 'title',
+                        'uri' => 'uri',
+                    ],
+                    'parents' => [],
+                ],
+                new Annotation(
+                    'id',
+                    "Nested list:\n\n- Item 1\n- Item 2\n  - Item 2.1\n    1. Item 2.1.1",
```
giorgiosironi: wonder if there's any format that would make it easier to provide many tests like these. Separating the data into text files may introduce too much indirection, but the `\n` impact readability

giorgiosironi: Heredoc is probably uglier. What about:
```
                    "Nested list:\n". 
                    "\n". 
                    "- Item 1\n". 
                    "- Item 2\n".
                    "  - Item 2.1\n".
                    "    1. Item 2.1.1" 
```
or even
```
                    $this->lines([
                        "Nested list:", 
                        "", 
                        "- Item 1", 
                        "- Item 2",
                        "  - Item 2.1\n",
                        "    1. Item 2.1.1",
                    ]) 
```
?

giorgiosironi:
The build fails because `symfony/filesystem` is pinned to `v4.0.3` in `composer.lock`, but that version requires PHP 7.1; rebuilding the `composer.lock` with 7.0 should solve that as it will switch to a lower version (it seems an indirect dependency).

src/Annotations/AppKernel.php
nlisgo: I could get rid of this, this is not required. I found it useful in development.

giorgiosironi: I found it useful too (to the point that I'd like it to be run also in other testing environments)

src/Annotations/Serializer/AnnotationNormalizer.php
nlisgo: Will refactor.

src/Annotations/Serializer/AnnotationNormalizer.php
nlisgo: We cannot absolutely rule it out. If we detect them I would like to be notified.

src/Annotations/Serializer/CommonMark/functions_include.php
nlisgo: I haven't experienced the issue myself but saw this:
https://github.com/guzzle/psr7/pull/22

src/HypothesisClient/Client/Search.php
nlisgo: The only values that should impact the total are `$username` and `$accessToken`. We do not hold on to access tokens. It is not necessary, just a performance optimisation.

src/Annotations/Serializer/CommonMark/functions_include.php
giorgiosironi: Not sure whether it's an issue for us or not; in theory, we should only include once the autoloader with `require_once 'vendor/autoload.php`, but that has to be verified.

src/HypothesisClient/Client/Search.php
giorgiosironi: Also time passing impacts the value e.g. if a process holds onto an instance of this class it doesn't see the count growing
So like all layers of caching I'd drop it until its need is measured (perhaps the API is so slow that we'll need another layer of caching anyway, or it's so fast that is unnecessary)

src/Annotations/Serializer/CommonMark/functions_include.php
nlisgo: I'll get rid of it. If we ever experience an issue we can reintroduce

src/Annotations/Serializer/AnnotationNormalizer.php
nlisgo: @NickDuf has approved the wording. I will ask him if he can provide different text for the 2 variants.

nlisgo:
I will remove the WIP when I have introduced appropriate sanity checks for the markdown.

Hypothesis use python `bleach` on their server app: https://github.com/hypothesis/h/blob/0e1516a17e575c47854504af53755ac9b80c28cf/h/util/markdown.py#L135-L137

and angular ngSanitize on the client app

config/test.php
```diff
@@ -3,5 +3,7 @@
 $config = require __DIR__.'/ci.php';
 
 $config['aws']['stub'] = true;
+$config['api_url'] = 'https://api.elifesciences.org/';
```
giorgiosironi: this is already in the formula for other environments :+1:

src/Annotations/Serializer/CommonMark/Block/Parser/MathMLParser.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace eLife\Annotations\Serializer\CommonMark\Block\Parser;
+
+use eLife\Annotations\Serializer\CommonMark\Block\Element\MathML;
+use League\CommonMark\Block\Parser\AbstractBlockParser;
+use League\CommonMark\ContextInterface;
+use League\CommonMark\Cursor;
+
+class MathMLParser extends AbstractBlockParser
```
giorgiosironi: would be helpful to document in the README where the architecture of these parsers come from, whether it is documentation or helpful Stack Overflow questions. A couple of links which will be very valuable when someone needs to make changes down the road

src/HypothesisClient/Serializer/AnnotationDenormalizer.php
```diff
@@ -0,0 +1,41 @@
...
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+final class AnnotationDenormalizer implements DenormalizerInterface, DenormalizerAwareInterface
+{
+    use DenormalizerAwareTrait;
+
+    public function denormalize($data, $class, $format = null, array $context = []) : Annotation
+    {
+        $data['document'] = $this->denormalizer->denormalize($data['document'], Document::class);
+        $data['target'] = $this->denormalizer->denormalize($data['target'][0], Target::class);
```
giorgiosironi: should we assert that `$data['target']` only has 1 element, or should we silently ignore any additional element in it?

tests/Annotations/AnnotationsTest.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        $client = static::createClient();
+
+        $this->mockHypothesisSearchCall('4321', $this->createAnnotations(), 20);
+        $client->request('GET', '/annotations?by=4321&access=restricted');
+        $response = $client->getResponse();
+        $this->assertResponseIsValid($response);
+
+        $this->assertSame(400, $response->getStatusCode());
+    }
```
giorgiosironi: I think there is one missing case: `it_will_return_public_only_annotations_to_unauthenticated_users`, targeting

- `/annotations?by=4321&access=public` (or `/annotations?by=4321` if the default access is `public`)
- mock is a call to hypothesis without an authentication token
- 200 response code

tests/Annotations/ApiTestCase.php
```diff
@@ -0,0 +1,248 @@
...
+                [],
+                json_encode($json)
+            )
+        );
+    }
+
+    final protected function createAnnotations($total = 10) : Traversable
+    {
+        for ($i = 1; $i <= $total; ++$i) {
+            // Allow a variety of annotation structures to be present, without being random.
```
giorgiosironi: I like that this is repeatable rather than random :+1:

tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -67,11 +70,12 @@ public function prepareDependencies()
         $this->secretKey = 'secret_key';
         $this->authority = 'authority';
         $this->authorization = sprintf('Basic %s', base64_encode($this->clientId.':'.$this->secretKey));
-        $this->credentials = new UserManagementCredentials('client_id', 'secret_key', 'authority');
+        $this->userManagement = new UserManagementCredentials('client_id', 'secret_key', 'authority');
+        $this->jwtSigning = new JWTSigningCredentials('client_id', 'secret_key', 'authority', new SystemClock());
```
giorgiosironi: can probably use `FixedClock` in this test for improved repeatability?

tests/Annotations/InMemoryStorageAdapter.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace tests\eLife\Annotations;
+
+use Csa\GuzzleHttp\Middleware\Cache\Adapter\StorageAdapterInterface;
+use Csa\GuzzleHttp\Middleware\Cache\CacheMiddleware;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class InMemoryStorageAdapter implements StorageAdapterInterface
```
giorgiosironi: a pity that this is not provided by https://github.com/csarrazi/guzzle-cache-middleware/tree/master/src/Adapter
I see no other choice than bundling it here for now :+1:

tests/HypothesisClient/Serializer/Annotation/Target/Selector/TextQuoteDenormalizerTest.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        $this->assertEquals($expected, $this->normalizer->denormalize($json, TextQuote::class));
+    }
+
+    public function denormalizeProvider() : array
+    {
+        return [
+            'complete' => [
+                [
+                    'exact' => 'exact',
```
giorgiosironi: Not a big deal, but in general in these data-driven tests it's helpful to differentiate values that are used differently, e.g.
````
[
    'exact' => 'e',
    'prefix' => 'p',
    'suffix' => 's',
]
// ...
new TextQuote('e', 'p', 's')
```
so that the relation between values is clear at a glance. Even better would be to have values taken from a real annotation for the `complete` samples.

giorgiosironi:
I can help and do some of my proposed changes, especially on the test side, as a PR towards this branch?

giorgiosironi:
All my changes are at https://github.com/elifesciences/annotations/pull/43

giorgiosironi:
> I will remove the WIP when I have introduced appropriate sanity checks for the markdown.

Since this is already huge, it would be interesting to slice off sanity checks and merge the API at it stands. The pages where this content would be displayed (or annotations would be created) are not publicly available so there is no risk.

nlisgo:
It has been decided that we will address the sanity checks in a subsequent pull request. Since we have editorial control we can hide any problem annotations in the mean time but I expect that we will have the sanity checks in place before we lift the feature flag.

giorgiosironi:
Indeed, "deployed in production" != "launched for real users"

PR #48, opened 2018-01-22
Latest Docker image should always be stable

giorgiosironi:
These tags should be pushed:

- `elifesciences/annotations:$commit` as soon as it is built on the mainline (not doing this for pull requests at the moment, to avoid creating too many tags)
- `elifesciences/annotations:approved` after the image has passed all `ci` and `end2end` tests
- `elifesciences/annotations:latest` during deploy in `prod`, which should follow shortly after. This is the default tag that should be checked out by new VMs (which are still the way we deploy in `end2end` and `prod`).

### DIFF
--- a/Jenkinsfile.prod
+++ b/Jenkinsfile.prod
@@ -8,6 +8,13 @@ elifePipeline {
     stage 'Deploy to prod', {
         elifeDeploySlackNotification 'annotations', 'prod'
         elifeGitMoveToBranch commit, 'master'
+        elifeOnNode(
+            {
+                sh "docker tag elifesciences/annotations_cli:${commit} elifesciences/annotations_cli:latest && docker push elifesciences/annotations_cli:latest"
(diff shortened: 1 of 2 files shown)


Jenkinsfile.prod
```diff
@@ -8,6 +8,13 @@ elifePipeline {
     stage 'Deploy to prod', {
         elifeDeploySlackNotification 'annotations', 'prod'
         elifeGitMoveToBranch commit, 'master'
+        elifeOnNode(
+            {
+                sh "docker tag elifesciences/annotations_cli:${commit} elifesciences/annotations_cli:latest && docker push elifesciences/annotations_cli:latest"
+                sh "docker tag elifesciences/annotations_fpm:${commit} elifesciences/annotations_fpm:latest && docker push elifesciences/annotations_fpm:latest"
```
thewilkybarkid: Is `latest` standard Docker terminology? Or could we use `master` to match the branch name?

giorgiosironi: It is a convention derived by the default tag used by `docker pull`:
https://hub.docker.com/r/jenkins/jenkins/tags/
https://docs.docker.com/engine/reference/commandline/pull/#examples

PR #50, opened 2018-01-23
Starting changing parents to ancestors

thewilkybarkid:
Implements https://github.com/elifesciences/api-raml/pull/211.

(Also implements https://github.com/elifesciences/api-raml/pull/209.)


tests/Annotations/Serializer/HypothesisClientAnnotationNormalizerTest.php
```diff
@@ -179,7 +183,6 @@ public function normalizeProvider() : array
                         'title' => 'title',
                         'uri' => 'uri',
                     ],
-                    'parents' => [],
```
giorgiosironi: to be clear, this is unrelated because the field is optional?

thewilkybarkid: Yep.

PR #52, opened 2018-01-23
Sanitize annotations

nlisgo:


nlisgo:
I believe that I am experiencing this issue: https://github.com/ezyang/htmlpurifier/issues/71

src/Annotations/AppKernel.php
```diff
@@ -103,6 +104,7 @@ public function __construct(string $environment = 'dev')
                 'authority' => '',
                 'group' => '',
             ],
+            'html_purifier.cache_dir' => $config['html_purifier']['cache_dir'] ?? __DIR__.'/../../var/html_purifier/cache',
```
nlisgo: We need to create the `/srv/annotations/var/html_purifier/cache` in all environments in order for this to pass.

src/Annotations/AppKernel.php
```diff
@@ -103,6 +104,12 @@ public function __construct(string $environment = 'dev')
                 'authority' => '',
                 'group' => '',
             ],
+            'html_purifier' => ($config['html_purifier'] ?? []) + [
+                'Cache.SerializerPath' => __DIR__.'/../../var/cache/html_purifier',
```
nlisgo: We need to create this directory in all environments: `/srv/annotations/var/cache/html_purifier`.

giorgiosironi: Is the `Directory /srv/annotations/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer not writable, please chmod to 777` error related to this?

nlisgo: It's related to that. I will work on `MarkdownSanitizerTest` so that it doesn't attempt to write to cache.

giorgiosironi: https://github.com/elifesciences/annotations/pull/54

src/Annotations/Serializer/CommonMark/Block/Renderer/LatexRenderer.php
```diff
@@ -1,26 +0,0 @@
-<?php
```
giorgiosironi: nice to see these classes simplified as function callbacks

src/Annotations/Serializer/CommonMark/MarkdownSanitizer.php
```diff
@@ -0,0 +1,29 @@
...
+    private $markdownConverter;
+
+    public function __construct(Converter $markdownConvertor, HtmlConverter $htmlConverter, HTMLPurifier $htmlPurifier)
+    {
+        $this->markdownConverter = $markdownConvertor;
+        $this->htmlConverter = $htmlConverter;
+        $this->htmlPurifier = $htmlPurifier;
+    }
+
+    public function parse(string $input) : string
```
giorgiosironi: This API conversion pipeline needs its step to be listed as documentation in the README:
```
1. Hypothesis API to `Model` objects
2. Denormalization
2.1 Conversion to HTML (uses library X)
2.2 HTML purification (uses library Y)
...
```
(ADRs are not a good format for that)

giorgiosironi: The obvious questions that needs to be answered there is why we sanitize intermediate HTML and convert back to Markdown rather than sanitizing the final HTML

composer.json
```diff
@@ -29,7 +29,6 @@
         "guzzlehttp/guzzle": "^6.3",
         "knplabs/console-service-provider": "^2.1",
         "league/commonmark": "^0.17.0",
-        "psr/container": "^1.0",
```
thewilkybarkid: Unrelated to this PR, but it _should_ be included since `AppKernel` implements `ContainerInterface`.

src/Annotations/AppKernel.php
```diff
@@ -103,6 +102,15 @@ public function __construct(string $environment = 'dev')
                 'authority' => '',
                 'group' => '',
             ],
+            'common_mark.environment' => ($config['common_mark']['environment'] ?? []) + [
+                'allow_unsafe_links' => false,
+            ],
+            'html_purifier' => ($config['html_purifier'] ?? []) + [
+                'Cache.SerializerPath' => __DIR__.'/../../var/cache/html_purifier',
```
thewilkybarkid: Could we extract the cache path as its own parameter?

README.md
```diff
@@ -43,3 +43,10 @@ docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh
 ```
 
 `-u` is needed to write to `vendor/`. Currently Composer prints some warning when git-cloning due to the user not being in /etc/passwd.
+
+## Journey of a request to the annotations api
+1. Query received via the `annotations` api.
+1. Retrieve annotation listing from `hypothes.is/api` based on query parameters.
+1. Denormalize the annotations into an object model.
```
giorgiosironi: [[HypothesisClient\Model]`?

src/Annotations/Serializer/CommonMark/functions.php
```diff
@@ -2,9 +2,25 @@
...
+use League\CommonMark\Util\Xml;
+
 function clean_paragraph($text)
 {
     $allowed_tags = '<i><sub><sup><span><del><math><a><br><caption>';
 
     return strip_tags($text, $allowed_tags);
 }
+
+function escape_math($text)
```
thewilkybarkid: These should have types. (`function escape_math(string $text) : string`.)

README.md
```diff
@@ -43,3 +43,10 @@ docker-compose -f docker-compose.ci.yml run ci ./project_tests.sh
...
+1. Normalize the annotations to prepare the response which conforms to the eLife json schema. [HTMLPurifier](http://htmlpurifier.org/) is used to sanitise content values. We rely on [CommonMark](http://commonmark.thephpleague.com/) to step through the Markdown content and prepare the structure that we need.
```
giorgiosironi: would invert the two sentences: first we step through, and while stepping through we can sanitise?

src/Annotations/AppKernel.php
giorgiosironi: You mean `var/cache/` or `/srv/annotations/var/cache`?

src/Annotations/AppKernel.php
```diff
@@ -103,6 +102,15 @@ public function __construct(string $environment = 'dev')
...
+            ],
+            'html_converter' => ($config['html_converter'] ?? []) + [
```
giorgiosironi: I think this is a leftover to be removed?

src/Annotations/Serializer/CommonMark/functions.php
```diff
@@ -2,9 +2,25 @@
...
 {
     $allowed_tags = '<i><sub><sup><span><del><math><a><br><caption>';
 
     return strip_tags($text, $allowed_tags);
 }
+
+function escape_math(string $text) : string
+{
+    // Escape MathML.
+    $escaped = preg_replace_callback('~(?P<before><math[^>]*>)(?P<mathml>.*)(?P<after></math>)~s', function ($match) {
```
giorgiosironi: I wonder if the regex is greedy and what happens with multiple <math> tags. I'll ensure there is a test for that.

src/Annotations/Serializer/CommonMark/functions.php
```diff
@@ -2,9 +2,25 @@
...
+        return Xml::escape($match['before'].$match['mathml'].$match['after']);
+    }, $text);
+    // Escape LaTeX.
+    $escaped = preg_replace_callback('~(?P<before>\$\$)(?P<latex>.+)(?P<after>\$\$)~s', function ($match) {
+        return Xml::escape($match['before'].$match['latex'].$match['after']);
+    }, $escaped);
+
+    return $escaped;
+}
```
giorgiosironi: also, I'm not sure whether an "unescaping" is necessary

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
```diff
@@ -4,13 +4,15 @@
...
 use League\CommonMark\Block\Element;
 use League\CommonMark\DocParser;
 use League\CommonMark\ElementRendererInterface;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use function eLife\Annotations\Serializer\CommonMark\escape_math;
 
 final class HypothesisClientAnnotationNormalizer implements NormalizerInterface, NormalizerAwareInterface
```
thewilkybarkid: `NormalizerAwareInterface` not needed anymore?

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
```diff
@@ -169,4 +171,9 @@ public function supportsNormalization($data, $format = null) : bool
     {
         return $data instanceof Annotation;
     }
+
+    private function purify(string $text) : string
+    {
+        return $this->htmlPurifier->purify($text);
```
thewilkybarkid: I suspect we'll want to cache the results of these (IIRC the library doesn't do this).

This class is feeling a bit bloated now, knowing how to convert all the bits and also purify them. Can purification happen before/after?

src/Annotations/Serializer/CommonMark/functions.php
nlisgo: unescaping is not necessary as LaTeX and MathML are being presented as escaped in the api response.

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
nlisgo: Purification happening after is a possibility. I implemented a solution where it was happening before but it wasn't optimal.

In order for it to happen after the `HypothesisClientAnnotationNormalizer` would return an unsanitized response and then we would need a class so step through each item in the `content` array. So, I would need 2 places where the structure of the response was known. Perhaps we can have a chat about this tomorrow.

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
nlisgo: Will remove

src/Annotations/Serializer/CommonMark/functions.php
nlisgo: I will add a test.

src/Annotations/AppKernel.php
nlisgo: Will remove

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
nlisgo: I've done a refactor here which means I can remove purify from this class. See what you think.

src/Annotations/Serializer/CommonMark/functions.php
thewilkybarkid: Named groups aren't needed? Could be simplified to:

```php
function escape_math(string $text) : string
{
    $escape = function (array $matches) {
        return Xml::escape($matches[0]);
    };

    return preg_replace_callback_array([
        '~(<math[^>]*>.*?</math>)~s' => $escape,
        '~(\$\$.+?\$\$)~s' => $escape,
    ], $text);
}
```

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
```diff
@@ -145,8 +143,8 @@ private function processListBlock(Element\ListBlock $block)
                 foreach ($item->children() as $child) {
                     if ($child instanceof Element\ListBlock) {
                         $items[] = [$render($child)];
-                    } else {
-                        $items[] = $this->htmlRenderer->renderBlock($child);
+                    } elseif ($item = $this->htmlRenderer->renderBlock($child)) {
```
thewilkybarkid: Not introduced by this PR, but `processText()` has already rendered the whole list then thrown it away (in `processText()`) in favour of doing each item here?

src/Annotations/Serializer/CommonMark/HtmlRenderer.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace eLife\Annotations\Serializer\CommonMark;
+
+use HTMLPurifier;
+use League\CommonMark\Block\Element\AbstractBlock;
+use League\CommonMark\Environment;
+use League\CommonMark\HtmlRenderer as CommonMarkHtmlRenderer;
+
+class HtmlRenderer extends CommonMarkHtmlRenderer
```
thewilkybarkid: The name doesn't describe what it's doing (`HtmlPurifierRenderer`?). Better to use composition too.

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
nlisgo: The render of the list converts the whole list to HTML. We need to step through the list to separate out each item. I will see if I can improve this.

src/Annotations/Serializer/CommonMark/HtmlRenderer.php
thewilkybarkid: Misses out `renderInlines()` too?

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
thewilkybarkid: You just need to stop it rendering the whole thing if it's a list, it's wasted effort.

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
```diff
@@ -8,11 +8,11 @@
 use League\CommonMark\DocParser;
 use League\CommonMark\ElementRendererInterface;
 use Psr\Log\LoggerInterface;
-use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use function eLife\Annotations\Serializer\CommonMark\escape_math;
 
-final class HypothesisClientAnnotationNormalizer implements NormalizerInterface, NormalizerAwareInterface
+final class HypothesisClientAnnotationNormalizer implements NormalizerInterface
```
thewilkybarkid: The trait needs to go too.

composer.json
```diff
@@ -24,10 +24,12 @@
         "elife/content-negotiator": "^1.0",
         "elife/logging-sdk": "^1.0@dev",
         "elife/ping": "^1.0",
+        "ezyang/htmlpurifier": "dev-master",
```
thewilkybarkid: Are we relying on commits that haven't been tagged yet?

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
```diff
@@ -80,7 +80,7 @@ public function normalize($object, $format = null, array $context = []) : array
 
     private function processText(string $text) : array
     {
-        $blocks = $this->docParser->parse($text)->children();
+        $blocks = $this->docParser->parse(escape_math($text))->children();
```
thewilkybarkid: This means we're escaping Math elements before, then purifying after? This could also compose `HtmlRenderer`?

(Means more regexes, but we're going to be better off caching the whole thing anyway.)

src/Annotations/Serializer/CommonMark/functions.php
```diff
@@ -2,9 +2,25 @@
 
 namespace eLife\Annotations\Serializer\CommonMark;
 
-function clean_paragraph($text)
+use League\CommonMark\Util\Xml;
+
+function clean_paragraph(string $text) : string
 {
     $allowed_tags = '<i><sub><sup><span><del><math><a><br><caption>';
```
thewilkybarkid: What's the difference between `<math>` here and `escape_math()`?

Is it a requirement to drop MathML? Unlike LaTeX we can display it...

tests/Annotations/Serializer/HypothesisClientAnnotationNormalizerWebTest.php
```diff
@@ -4,31 +4,16 @@
...
-use PHPUnit_Framework_TestCase;
-use Symfony\Component\Debug\BufferingLogger;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use tests\eLife\Annotations\WebTestCase;
 
 /**
  * @covers \eLife\Annotations\Serializer\HypothesisClientAnnotationNormalizer
  */
-final class HypothesisClientAnnotationNormalizerTest extends PHPUnit_Framework_TestCase
+final class HypothesisClientAnnotationNormalizerWebTest extends WebTestCase
```
thewilkybarkid: `WebTest`? Should just extend `ApplicationTestCase` if you want `AppKernel`.

src/Annotations/Serializer/CommonMark/HtmlRenderer.php
nlisgo: I was working under that assumption. I will verify it by introducing a test.

src/Annotations/Serializer/CommonMark/functions.php
nlisgo: Not a requirement. I think it will not be too much trouble do capture and display it particularly when it's in a block rather than inline. But I will take that on as a followup to this. `<math>` was left in while I was deciding how to treat mathml. I will remove it for now while we are simply escaping MathML.

src/Annotations/Serializer/HypothesisClientAnnotationNormalizer.php
nlisgo: The Math elements are escaped while the content is in Markdown. We need to determine what `CommonMark` `Block` elements that we have before rendering the HTML.

thewilkybarkid: You mean that `CommonMark` does something with HTML element? Assumed it wouldn't...

src/Annotations/Serializer/CommonMark/HtmlPurifierRenderer.php
```diff
@@ -0,0 +1,72 @@
...
+
+class HtmlPurifierRenderer implements ElementRendererInterface
+{
+    private $htmlPurifier;
+    protected $environment;
+
+    /**
+     * @param Environment $environment
+     */
+    public function __construct(Environment $environment, HTMLPurifier $htmlPurifier)
```
thewilkybarkid: I meant this should compose the original `HtmlRender` rather than duplicate everything...

src/Annotations/AppKernel.php
thewilkybarkid: Something like https://github.com/elifesciences/annotations/blob/83c91e5ea7e296cd52c1df0fec4c862431268388/src/Annotations/AppKernel.php#L81

src/Annotations/Serializer/CommonMark/HtmlPurifierRenderer.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public function renderInlines($inlines) : string
+    {
+        return $this->renderer->renderInlines($inlines);
+    }
+
+    public function renderBlock(AbstractBlock $block, $inTightList = false) : string
+    {
+        return strip_tags($this->htmlPurifier->purify($this->renderer->renderBlock($block, $inTightList)), self::ALLOW_TAGS);
```
thewilkybarkid: Would be easier to read over a few lines.

tests/Annotations/Serializer/HypothesisClientAnnotationNormalizerApplicationTest.php
```diff
@@ -738,6 +728,135 @@ public function normalizeProvider() : array
...
+                    ],
+                    'created' => $createdDate,
+                    'document' => [
+                        'title' => 'title',
+                        'uri' => 'uri',
+                    ],
+                ],
+                new Annotation(
+                    'id',
+                    '<script>evil()</script>',
```
thewilkybarkid: Since the code for purification etc has been extracted, these should be unit tests on the relevant renderer.

(Since this is important I'm happy with a single high-level test that also checks that it's configured correctly, but that doesn't belong here.)

src/Annotations/Serializer/CommonMark/Block/Renderer/CodeRenderer.php
```diff
@@ -17,6 +17,6 @@ public function render(AbstractBlock $block, ElementRendererInterface $htmlRende
             throw new \InvalidArgumentException('Incompatible block type: '.get_class($block));
         }
 
-        return Xml::escape($block->getStringContent());
+        return Xml::escape(trim($block->getStringContent()));
```
thewilkybarkid: Whitespace might be meaningful?

thewilkybarkid:
@nlisgo Are you rebasing this branch? Easier to do a merge (keeps the history in the PR intact). Since we're doing squash commits we don't need to worry about the extra commit created.

src/Annotations/AppKernel.php
```diff
@@ -375,20 +378,24 @@ public function __construct(string $environment = 'dev')
...
-        $this->app['annotation.serializer.common_mark.html_renderer'] = function () {
+        $this->app['annotation.serializer.common_mark.html_renderer'] = function (Application $app) {
             return new HtmlRenderer($this->app['annotation.serializer.common_mark.environment']);
         };
 
+        $this->app['annotation.serializer.html_purifier'] = function (Application $app) {
+            return new HTMLPurifier($app['html_purifier']);
+        };
+
+        $this->app['annotation.serializer.common_mark.math_escape_renderer'] = function () {
```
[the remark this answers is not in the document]
nlisgo: @thewilkybarkid Could you submit a PR for this or provide a code snippet? I see different uses for `extend`, I'm not sure what the appropriate one is here.

src/Annotations/AppKernel.php
```diff
@@ -375,20 +380,24 @@ public function __construct(string $environment = 'dev')
             return new DocParser($this->app['annotation.serializer.common_mark.environment']);
         };
 
-        $this->app['annotation.serializer.common_mark.html_renderer'] = function () {
+        $this->app['annotation.serializer.common_mark.html_renderer'] = function (Application $app) {
             return new HtmlRenderer($this->app['annotation.serializer.common_mark.environment']);
         };
 
+        $this->app['annotation.serializer.html_purifier'] = function (Application $app) {
+            return new HTMLPurifier($app['html_purifier']);
```
thewilkybarkid: Using `$this->app` elsewhere.

tests/Annotations/Serializer/CommonMark/HtmlPurifierRendererTest.php
```diff
@@ -26,7 +26,7 @@ public function setup()
         $this->renderer = $this->getMockBuilder(ElementRendererInterface::class)
             ->getMock();
         $this->abstractBlock = $this->createMock(AbstractBlock::class);
-        $this->htmlPurifierRenderer = new HtmlPurifierRenderer($this->renderer, new HTMLPurifier());
+        $this->htmlPurifierRenderer = new HtmlPurifierRenderer($this->renderer, new HTMLPurifier(['Cache.SerializerPath' => '/tmp']));
```
thewilkybarkid: `sys_get_temp_dir()`

PR #56, opened 2018-01-24
Create var/cache/html_purifier folder

giorgiosironi:
Also remove another folder that should be created by the base image now.


thewilkybarkid:
Because of a bug in the library that can't create the folder? Might be worth adding a note.

giorgiosironi:
Not sure it's a bug, many libraries assume that if you give them a path, they should use that path rather than creating the folder by themselves.

PR #57, opened 2018-01-24
Exposing problem with greedy regex

giorgiosironi:


giorgiosironi:
The test is probably not the best place to capture this, but I was unsure about unescaping to place it directly on the function. `<latex>` likely has the same issue.

PR #58, opened 2018-01-25
Extract class to deal with Docker images

giorgiosironi:


giorgiosironi:
Not much testing that can be performed on the branch, since it involves image tagging and pushing, which happens only on `develop`.

PR #62, opened 2018-01-25
[WIP] Cache access tokens

thewilkybarkid:
It seems that the access tokens are valid for an hour (TBC), which means that if someone's paginating through their own annotations on Journal there's a request to get a token every time (and it has to come before the main search request), slowing it down.

The token should be cached for its lifetime.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,9 @@
         "guzzlehttp/guzzle": "^6.3",
         "knplabs/console-service-provider": "^2.1",
         "league/commonmark": "^0.17.0",
+        "psr/simple-cache": "^1.0",
         "silex/silex": "^2.2",
+        "symfony/cache": "^4.0",
         "symfony/property-access": "^3.3",
         "symfony/serializer": "^3.3",
         "willdurand/negotiation": "^2.3"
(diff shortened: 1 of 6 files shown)


thewilkybarkid:
This first pass puts it in the `HypothesisClient` realm, but I'm thinking this should actually live inside the application.

If the token is revoked there's currently no way to fetch a new one... though that said, if it's been revoked that probably wouldn't work?

giorgiosironi:
Since `expires_in` is an approximation, we need to know what error can come out of using an expired token; that error should lead to asking for a new token, as the ultimate cache invalidation strategy.

nlisgo:
@giorgiosironi I will find out.

nlisgo:
Unfortunately, it appears that an expired token is simply ignored.

Response from @robertknight:
> It looks to me (based on `h/auth/policy.py` in h) like expired tokens are simply ignored. On that basis you’ll need to compute the expiry status yourself. The Hypothesis client uses `now() + token.expires_in - A_FEW_MINUTES` for this where `A_FEW_MINUTES` allows for the time taken to fetch the token in the first place.
> IMO that’s not very friendly and we ought to throw a useful exception if the token is invalid, but that’s the current status.

thewilkybarkid:
Ouch. Guess the same applies for revoked (if that's even happens).

Think that refers to https://github.com/hypothesis/client/blob/509e8f59dc98167c49b3e63ed9acffe6df1382f6/src/sidebar/util/oauth-client.js#L30 which is 10 seconds(?), but we at least want the timeout on the request (so maybe 1/2 minutes would be prudent).

PR #63, opened 2018-01-25
Compose api-dummy container

giorgiosironi:
Now that api-dummy is provided as a container, a pinned version can be composed in dev and ci for exploratory testing.

(other environments would use a real API)


config/container.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 return [
+    'api_url' => 'https://api_dummy:8080',
```
thewilkybarkid: `http`?

giorgiosironi: 👍

PR #64, opened 2018-01-26
[WIP] Configure Guzzle clients

thewilkybarkid:


src/Annotations/AppKernel.php
```diff
@@ -224,7 +224,10 @@ public function __construct(string $environment = 'dev')
 
             return new Client([
                 'base_uri' => $this->app['hypothesis']['api_url'],
+                'connect_timeout' => 0.5,
+                'decode_content' => 'gzip',
                 'handler' => $this->app['hypothesis.guzzle.handler'],
+                'timeout' => 0.9,
```
thewilkybarkid: We want fast failures as Journal has a 1-second timeout. This has to be less than that otherwise Annotations won't see an error (Journal will just have disconnected).

src/Annotations/AppKernel.php
```diff
@@ -224,7 +224,10 @@ public function __construct(string $environment = 'dev')
 
             return new Client([
                 'base_uri' => $this->app['hypothesis']['api_url'],
+                'connect_timeout' => 0.5,
+                'decode_content' => 'gzip',
```
thewilkybarkid: Probably worth doing as this is external.

thewilkybarkid:
We might want to use HTTP/2 for Hypothesis (their API supports it), but not sure what to make of https://github.com/guzzle/guzzle/issues/1249 (means upgrading cURL anyhow).

src/Annotations/AppKernel.php
thewilkybarkid: Just confirmed that the current hypothesis-dummy supports this for testing (happens at the cURL level, which the local tests don't hit), though https://github.com/elifesciences/hypothesis-dummy/pull/3 will turn it off (removes Nginx; could add it to the PHP code?).

giorgiosironi: Should degrade automatically if the response is not gzipped as there's no `Content-Encoding` header? Which would be a difference between `end2end` and `prod` but not critical?

thewilkybarkid: Yep, cURL handles it all.

giorgiosironi: As it stands, this should not explode with any version of `hypothesis-dummy` then

src/Annotations/AppKernel.php
giorgiosironi: It's reasonable to think that the rest of the code being executed by this service will have a 100ms overhead

PR #66, opened 2018-01-29
Containerized proofreader setup

giorgiosironi:
We now have a faster and more stable setup for proofreader, which copies it from another prebuilt image.


giorgiosironi:
@nlisgo getting this into `develop`

PR #67, opened 2018-01-29
Pin base images

giorgiosironi:
There is one obvious and one subtle reason for doing this.

## Obvious: build reproducibility.

If you start from the same commit, the
same image should be produced, which fails or passes the same tests. We shouldn't rely on the default `latest`
as this can change over time quite a lot. This is an image fully under
our control.

## Subtle: new images 

New images are generated as `elifesciences/annotations_cli:$my_commit` where the SHA `$my_commit` is
the tag (e.g. `elifesciences/annotations_cli:0da352bbfe3f431f7193bc734757c43d455e4c8f`). If we have a new base image, we can't deploy its update without also adding an empty commit to the `annotations` repository. Why? Because that tag (much like a `git tag`) will be cached everywhere, preventing nodes from getting the new image unless they explicitly pull it again. Essentially this is `composer.lock` over the base image rather than always depending on `dev-master`.

## Automated updates

Like for all pinnings, automated update applies: we can have pipelines
proposing updates with newly released base images.

## Tag names

Suggestions are welcome on how to tag base images in a more readable
way:

- time of release (current, gives an idea of how old it is, but to be
unambiguous has to include minutes and seconds)
- commit (even more unreadable)
- manually generated release tag (like `git tag`)
- some metadata included in a `version.txt|json|yaml|doc|pdf` file


thewilkybarkid:
Big 👍 

I'd say commits are the 'correct' way to tag given how we work (semver/Git tags would be ideal, but has implications). Agree they're unreadable, but since there'll be an automated process to update them, add a readable timestamp as a comment after? (Commit message too might be useful too?)

giorgiosironi:
I'll get this in with these tags, then go onto `elife-base-images` to write an ADR for this along with the implementation of the tags.

PR #71, opened 2018-02-02
Ship smoke tests with images

giorgiosironi:
The `fpm` and `cli` images are the ones to be used in production. Both now ship their own smoke tests, respectively checking `/ping` through the FastCGI protocol and checking `bin/console` can be run. The build runs all containers together and then executes an additional bash script in each of `fpm` and `cli` that runs the correct smoke tests.

Incidentally, we only need a `web/app.php` folder now, as it can load the correct environment using `ENVIRONMENT_NAME`.

The `PROJECT_FOLDER` environment variable is introduced with a default of `/srv/annotations`. This variable name should become a standard and is lifted from `builder-base-formula` terminology.

Some parts about the FastCGI smoke testing can be moved to the base image for reuse, once integrated in this project.


thewilkybarkid:
The smoke tests as-is are web-only, so only make sense for `fpm`. That said, you're right that they should both have smoke tests, so `cli` should have a separate set that test `bin/console`.

giorgiosironi:
In an ideal world we would define `HEALTHCHECK` in the container image, but we're not ready to take advantage of that metadata yet. I'll try to replicate these checks into the two container images.

giorgiosironi:
(Notice the image is `fpm`, not `web` so it needs an nginx somewhere to be accessed and can't just do local tests)

smoke_tests_fpm.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+# TODO: we should need only one web/app.php file?
+# use $ENVIRONMENT_NAME with a default?
+ping=$(SCRIPT_FILENAME=/srv/annotations/web/app_dev.php/ping REQUEST_URI=/ping REQUEST_METHOD=GET cgi-fcgi -bind -connect 127.0.0.1:9000 | tail -n 1)
```
giorgiosironi: most of this should go into the base image too, but it's nice to be able to smoke test `php-fpm` and the application without nginx

smoke_tests_fpm.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+# TODO: we should need only one web/app.php file?
+# use $ENVIRONMENT_NAME with a default?
```
thewilkybarkid: ELPP-3367 might well head that way.

smoke_tests_fpm.sh
thewilkybarkid: 👍

thewilkybarkid: Can we make the `SCRIPT_FILENAME ` dynamic to avoid knowledge of `/srv/annotations` here?

smoke_tests_fpm.sh
giorgiosironi: I think it can be done right now in this service, because the only difference between is a string containing `$ENVIRONMENT_NAME`

giorgiosironi:
Build needs some debugging, probably related to stdin and other streams being different in the Jenkins headless environment.

smoke_tests_fpm.sh
giorgiosironi: Could introduce a `PROJECT_FOLDER` environment variable (as explicit arguments would make it difficult to use)

giorgiosironi:
Updated title&description, ready for review.

web/app.php
```diff
@@ -5,7 +5,7 @@
 
 require_once __DIR__.'/../vendor/autoload.php';
 
-$app = new AppKernel('ci');
+$app = new AppKernel(getenv('ENVIRONMENT_NAME'));
```
thewilkybarkid: Needs a default of `dev`?

smoke_tests_fpm.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+ping=$(curl_fpm /ping | tail -n 1)
+echo "GET /ping: $ping"
+[ "$ping" == "pong" ]
+
```
thewilkybarkid: Extra line

curl_fpm
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -e
+# TODO: this script must be moved to the base fpm image
```
thewilkybarkid: 👍

thewilkybarkid:
`smoke.sh` isn't used anymore?

web/app.php
giorgiosironi: Can't hurt, although it should be provided by `dev.env`

giorgiosironi:
> smoke.sh isn't used anymore?

That works against HTTP, while here we have `fpm` and `cli` so it can't be used in this project without nginx. Remains to be seen if it will still be used to test a whole project (in integration with the nginx coming from the formula).

thewilkybarkid:
So can be removed from `Dockerfile.ci`?

Dockerfile.fpm
```diff
@@ -11,10 +15,10 @@ RUN composer install
 # yes this is how you copy directories
```
thewilkybarkid: Can't comment on the line above this, but the `TODO` can be removed.

Dockerfile.fpm
```diff
@@ -11,10 +15,10 @@ RUN composer install
...
 COPY bin/ /srv/annotations/bin/
 COPY config/ /srv/annotations/config/
-COPY phpunit.xml.dist /srv/annotations/
 COPY src/ /srv/annotations/src/
 COPY tests/ /srv/annotations/tests/
```
thewilkybarkid: Not needed if `phpunit.xml.dist` is being removed?

Dockerfile.fpm
```diff
@@ -1,7 +1,11 @@
...
 USER root
 RUN docker-php-ext-install pcntl
+# TODO: move in base image
+RUN apt-get update && apt-get install libfcgi0ldbl
+COPY curl_fpm /srv/bin/
 
+ENV PROJECT_FOLDER=/srv/annotations
 WORKDIR /srv/annotations
 COPY composer.json composer.lock /srv/annotations/
 # could probably use --no-dev and other optimizations?
```
thewilkybarkid: If PHPUnit is not being configured, add `--no-dev`.

Dockerfile.fpm
giorgiosironi: I think the `Dockerfile.ci` file could do much of the tests set up in place of this

thewilkybarkid: 👍  Uses `Dockerfile.cli` so definitely not needed here, but `Dockerfile.ci` should add `tests/` itself anyway (so CLI and FPM match).

Dockerfile.fpm
```diff
@@ -1,20 +1,22 @@
...
+COPY curl_fpm /srv/bin/
 
+ENV PROJECT_FOLDER=/srv/annotations
 WORKDIR /srv/annotations
 COPY composer.json composer.lock /srv/annotations/
-# could probably use --no-dev and other optimizations?
-RUN composer install
+RUN composer --no-interaction install --no-suggest --no-dev
 
 # TODO: remove phpunit.xml.dist from here?
```
thewilkybarkid: Can be removed.

Dockerfile.fpm
```diff
@@ -1,20 +1,21 @@
...
+# TODO: move in base image
+RUN apt-get update && apt-get install libfcgi0ldbl
+COPY curl_fpm /srv/bin/
 
+ENV PROJECT_FOLDER=/srv/annotations
 WORKDIR /srv/annotations
 COPY composer.json composer.lock /srv/annotations/
-# could probably use --no-dev and other optimizations?
-RUN composer install
+RUN composer --no-interaction install --no-suggest --no-dev
```
thewilkybarkid: `--classmap-authoritative` as well?

giorgiosironi: It may not break `dev`, because https://github.com/elifesciences/annotations/blob/develop/docker-compose.yml#L20 will mount the whole folder (including `vendor/`) inside the container.

Dockerfile.ci
```diff
@@ -3,16 +3,17 @@ FROM elifesciences/proofreader-php:latest AS proofreader
...
-RUN git clone https://github.com/asm89/smoke.sh /opt/smoke.sh
-
 RUN mkdir -p build && chown www-data:www-data build
 
 USER elife
 COPY --from=proofreader --chown=elife:elife /srv/proofreader-php /srv/proofreader-php
 RUN ln -s /srv/proofreader-php/bin/proofreader /srv/bin/proofreader
 
+USER root
+RUN composer --no-interaction install --no-suggest
```
thewilkybarkid: CI should be fine with `--classmap-authoritative` too, makes it a bit faster.

Dockerfile.ci
```diff
@@ -3,16 +3,17 @@ FROM elifesciences/proofreader-php:latest AS proofreader
...
-# extract into another image
-RUN git clone https://github.com/asm89/smoke.sh /opt/smoke.sh
-
 RUN mkdir -p build && chown www-data:www-data build
 
 USER elife
 COPY --from=proofreader --chown=elife:elife /srv/proofreader-php /srv/proofreader-php
 RUN ln -s /srv/proofreader-php/bin/proofreader /srv/bin/proofreader
 
+USER root
```
thewilkybarkid: Hmm, Composer should be run as `elife`?

Dockerfile.ci
```diff
@@ -3,16 +3,17 @@ FROM elifesciences/proofreader-php:latest AS proofreader
...
+RUN composer --no-interaction install --no-suggest --classmap-authoritative
+
 USER www-data
```
thewilkybarkid: Files should be `elife`?

giorgiosironi: Still experimental, but seemed to work in `hypothesis-dummy` https://github.com/elifesciences/hypothesis-dummy/blob/master/Dockerfile#L3-L8
Would change it as a separate PR because it influences all the `COPY` instructions which need to specify explicitly the owner, which would overly confuse the diff of the `Dockerfile`s here

Dockerfile.ci
giorgiosironi: Base test classes seem not being picked up 
```
Fatal error: Class 'tests\eLife\Annotations\WebTestCase' not found in /srv/annotations/tests/Annotations/AnnotationsTest.php on line 10
script returned exit code 255
```
from https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fannotations/detail/PR-71/13/pipeline

giorgiosironi: Can reproduce locally `docker-compose -f docker-compose.ci.yml run --rm ci ./project_tests.sh`

giorgiosironi: Seems to be missing from the class map:
```
www-data@1096fc52059f:/srv/annotations$ grep -r WebTestCase vendor/composer/    
vendor/composer/autoload_static.php:        'Silex\\WebTestCase' => __DIR__ . '/..' . '/silex/silex/src/Silex/WebTestCase.php',
vendor/composer/autoload_classmap.php:    'Silex\\WebTestCase' => $vendorDir . '/silex/silex/src/Silex/WebTestCase.php',
```

giorgiosironi: Regenerating manually the class map sees it appear:
```
root@9ea8d6e777ad:/srv/annotations# grep -r WebTestCase vendor/composer/
vendor/composer/autoload_static.php:        'Silex\\WebTestCase' => __DIR__ . '/..' . '/silex/silex/src/Silex/WebTestCase.php',
vendor/composer/autoload_static.php:        'tests\\eLife\\Annotations\\WebTestCase' => __DIR__ . '/../..' . '/tests/Annotations/WebTestCase.php',
vendor/composer/autoload_classmap.php:    'Silex\\WebTestCase' => $vendorDir . '/silex/silex/src/Silex/WebTestCase.php',
vendor/composer/autoload_classmap.php:    'tests\\eLife\\Annotations\\WebTestCase' => $baseDir . '/tests/Annotations/WebTestCase.php',
```

Dockerfile.ci
```diff
@@ -3,16 +3,18 @@ FROM elifesciences/proofreader-php:latest AS proofreader
...
 RUN mkdir -p build && chown www-data:www-data build
 
 USER elife
 COPY --from=proofreader --chown=elife:elife /srv/proofreader-php /srv/proofreader-php
 RUN ln -s /srv/proofreader-php/bin/proofreader /srv/bin/proofreader
 
-USER www-data
-COPY .php_cs project_tests.sh smoke_tests.sh /srv/annotations/
+USER root
+RUN composer   --no-interaction install --no-suggest --classmap-authoritative
```
thewilkybarkid: Can do `--no-autoloader`  if later doing a `composer dump-autoload`.

Dockerfile.cli
```diff
@@ -3,20 +3,18 @@ FROM elifesciences/php_cli:20180129144044
...
-# TODO: remove phpunit.xml.dist from here?
 # yes this is how you copy directories
 COPY bin/ /srv/annotations/bin/
 COPY config/ /srv/annotations/config/
-COPY phpunit.xml.dist /srv/annotations/
 COPY src/ /srv/annotations/src/
-COPY tests/ /srv/annotations/tests/
 COPY web/ /srv/annotations/web/
+COPY smoke_tests_cli.sh /srv/annotations/
+RUN composer dump-autoload --classmap-authoritative
```
thewilkybarkid: `--no-dev`

Dockerfile.fpm
```diff
@@ -1,20 +1,22 @@
...
-# TODO: remove phpunit.xml.dist from here?
 # yes this is how you copy directories
 COPY bin/ /srv/annotations/bin/
 COPY config/ /srv/annotations/config/
-COPY phpunit.xml.dist /srv/annotations/
 COPY src/ /srv/annotations/src/
-COPY tests/ /srv/annotations/tests/
 COPY web/ /srv/annotations/web/
+COPY smoke_tests_fpm.sh /srv/annotations/
+RUN composer dump-autoload --classmap-authoritative
```
thewilkybarkid: `--no-dev`

Dockerfile.ci
```diff
@@ -3,16 +3,18 @@ FROM elifesciences/proofreader-php:latest AS proofreader
...
 RUN ln -s /srv/proofreader-php/bin/proofreader /srv/bin/proofreader
 
-USER www-data
-COPY .php_cs project_tests.sh smoke_tests.sh /srv/annotations/
+USER root
+RUN composer --no-interaction install --no-suggest --no-autoloader
 
-CMD ["/bin/bash", "project_tests.sh"]
+COPY tests/ /srv/annotations/tests/
+RUN composer dump-autoload --classmap-authoritative
```
thewilkybarkid: Should probably the last provisioning step?

giorgiosironi: The sooner it is done, the more it can be cached; but in the case of `ci` it doesn't matter a great deal, as it is almost always being entirely re-executed (except for replayed builds)

thewilkybarkid: Thought it to be _less_ cacheable than the below files, but I guess if the `src` or `vendor` has changed anyway nothing's cached anymore so doesn't matter...

giorgiosironi: Going for clarity, I'll put the `COPY` together, pushing the autoload down as the last step that is executed as `root` (later on will be executed as `elife`).

PR #72, opened 2018-02-05
Make elife user the owner of the application

giorgiosironi:


Dockerfile.fpm
```diff
@@ -1,25 +1,30 @@
 FROM elifesciences/php_fpm:20180129144019
+
 USER root
 RUN docker-php-ext-install pcntl
 # TODO: move in base image
 RUN apt-get update && apt-get install libfcgi0ldbl
 COPY curl_fpm /srv/bin/
```
thewilkybarkid: `elife`?

giorgiosironi: Yes, no reason to keep it `root`

giorgiosironi: (will soon be removed from here)

PR #74, opened 2018-02-06
Update base image to first SHA-based tags

giorgiosironi:


giorgiosironi:
First builds should fail requiring `PHP_ENTRYPOINT`.

giorgiosironi:
Re-updated this with new infrastructure provided by `elife-base-images`.

PR #76, opened 2018-03-01
Code tidy

thewilkybarkid:


thewilkybarkid:
Probably want to mount `.php_cs.cache` when running Proofreader so that it's cached between builds.

docker-compose.ci.yml
```diff
@@ -41,6 +41,7 @@ services:
         volumes:
             - ./config/container.php:/srv/annotations/config.php
             - logs:/srv/annotations/var/logs
+            - .php_cs.cache:/srv/annotations/.php_cs.cache
```
giorgiosironi: so, this is a path that is local to the Jenkins workspace (local clone) so it may be kept around between builds of the same branch but not necessarily across pull requests (if that's needed). Per-project absolute path could be better?

giorgiosironi: also, the `ci` container is the one used to run project tests (which includes `proofreader`/`php-cs-fixer`) so it should be mounted there

docker-compose.ci.yml
```diff
@@ -75,6 +75,7 @@ services:
         volumes:
             - ./build:/srv/annotations/build
             - ./config/container.php:/srv/annotations/config.php
+            - .php_cs.cache:/srv/annotations/.php_cs.cache
```
giorgiosironi: Can't reproduce the failure locally yet, there may be some state in the workspace. I suspect however that if this is a single file, when it does not exist it is created as an empty folder which is not what we want?

giorgiosironi: My local build stops at
```
PHP-CS-Fixer consistency check with local project: /srv/proofreader-php/.php_cs vs .php_cs
Discrepancy between /srv/proofreader-php/.php_cs and .php_cs: cannot continue
--- Original
+++ New
@@ @@
   ),
-   'usingCache' => false,
+   'usingCache' => true,
    'usingLinter' => true,
    'hideProgress' => false,
 ))
```

giorgiosironi: ah, the `elife-libraries--ci` server must be using an old version of the `proofreader` image. Interesting

giorgiosironi: (or I must be)

giorgiosironi: But if we do `docker-compose build --pull` the `ci` image fails to build because it attempts to pull `elifesciences/annotations_cli:$commit` to extend it; but we don't push all images to Docker Hub. Better to pin the `proofreader-php` image.

giorgiosironi: Finally I get the same result. Now, is there a way to move the `.php_cs.cache` location so that we can mount a folder rather than a file that usually does not exist yet?

thewilkybarkid: Inside the container? We'll need to move to PHP-CS-Fixer 2.

giorgiosironi: Brought up https://elifesciences.atlassian.net/browse/ELPP-3163 in priority but this PR should still go ahead without the cache to get the rest of the benefits.

giorgiosironi:
Is this ready for review or does the `[WIP]` still apply?

thewilkybarkid:
Probably more to look at, but it's getting a little unwieldy so let's think about merging.

src/Annotations/Serializer/CommonMark/Block/Renderer/BlockQuoteRenderer.php
```diff
@@ -7,7 +7,7 @@
 use League\CommonMark\Block\Renderer\BlockRendererInterface;
 use League\CommonMark\ElementRendererInterface;
 
-class BlockQuoteRenderer implements BlockRendererInterface
+final class BlockQuoteRenderer implements BlockRendererInterface
```
giorgiosironi: maybe the language should introduce a `extendable` keyword (will never happen)

src/HypothesisClient/ApiSdk.php
```diff
@@ -20,21 +20,13 @@
 
 final class ApiSdk
 {
-    /** @var string */
```
giorgiosironi: were these IDE-related?

src/HypothesisClient/Model/Annotation.php
```diff
@@ -22,26 +22,16 @@
...
         string $id,
-        $text,
+        string $text = null,
         DateTimeImmutable $created,
         DateTimeImmutable $updated,
         Document $document,
         Target $target,
         string $uri,
-        $references,
+        array $references,
```
giorgiosironi: this was nullable in the API (missing field), so I'll check in the Serializer how that value is dealt with

src/HypothesisClient/Serializer/AnnotationDenormalizer.php
```diff
@@ -31,7 +31,7 @@ public function denormalize($data, $class, $format = null, array $context = [])
             $data['document'],
             $data['target'],
             $data['uri'],
-            $data['references'] ?? null,
+            $data['references'] ?? [],
```
giorgiosironi: it is dealt with like this, ok. We don't implement `normalize`

tests/Annotations/Command/QueueImportCommandTest.php
```diff
@@ -21,7 +21,7 @@
 /**
  * @covers \eLife\Annotations\Command\QueueImportCommand
  */
-class QueueImportCommandTest extends PHPUnit_Framework_TestCase
+final class QueueImportCommandTest extends PHPUnit_Framework_TestCase
```
giorgiosironi: is `final` for tests something that happens in other projects too? We have to come up with a check for `proofreader` then

tests/HypothesisClient/Client/TokenTest.php
```diff
@@ -42,9 +43,7 @@ public function prepareDependencies()
         $this->clientSecret = 'client_secret';
         $this->authority = 'authority';
         $this->group = 'group';
-        $this->credentials = $this->getMockBuilder(JWTSigningCredentials::class)
-            ->setConstructorArgs([$this->clientId, $this->clientSecret, $this->authority, new SystemClock()])
-            ->getMock();
+        $this->credentials = new JWTSigningCredentials($this->clientId, $this->clientSecret, $this->authority, new FixedClock());
```
giorgiosironi: this setup looks like something for a future trait or base class

PR #77, opened 2018-03-05
Remove mocking

thewilkybarkid:
(Extract some from #76.)


tests/HypothesisClient/Exception/ApiExceptionTest.php
```diff
@@ -18,13 +17,6 @@ class ApiExceptionTest extends PHPUnit_Framework_TestCase
      */
     public function it_requires_a_message()
     {
-        try {
-            $this->getMockBuilder(ApiException::class)->getMock();
-            $this->fail('A message is required');
-        } catch (TypeError $error) {
-            $this->assertTrue(true, 'A message is required');
-            $this->assertContains('must be of the type string', $error->getMessage());
-        }
```
thewilkybarkid: These seem to be testing a type hint, which wouldn't need to go through a mock anyway.

tests/HypothesisClient/Exception/ApiExceptionTest.php
```diff
@@ -34,7 +26,7 @@ public function it_requires_a_message()
      */
     public function it_has_an_error_code_of_zero()
     {
-        $e = $this->createMock(ApiException::class);
+        $e = new ApiException('foo');
         $this->assertEquals(0, $e->getCode());
```
thewilkybarkid: Assertion on a mock. 🙃

giorgiosironi: If this was `assertSame` it would have correctly failed due to the mocked method returning `null`?

tests/HypothesisClient/Exception/HttpProblemTest.php
```diff
@@ -1,89 +0,0 @@
-<?php
```
giorgiosironi: original class is abstract so ok to test it through its concrete subclasses

tests/HypothesisClient/Exception/ApiExceptionTest.php
thewilkybarkid: Apparently not. 🤔

`Exception::getCode()` is `final`, so must be a default.

thewilkybarkid: But yes, should be `assertSame` anyway.

PR #79, opened 2018-03-08
ELPP-3163 Proofreader upgrade to 0.2

giorgiosironi:
How https://github.com/elifesciences/proofreader-php/pull/7 can be used (depends on that of course, but can be built locally by rebuilding the `elifesciences/proofreader-php` image).

Supports this complex command to run the proofreader image on the host's files:
```
docker run -v $(pwd):/code -v $(pwd)/.php_cs.cache:/srv/proofreader-php/.php_cs.cache -u $(id -u) elifesciences/proofreader-php vendor/bin/php-cs-fixer fix /code/src
```

This will conflict with other PRs, so it's a proof of concept, not meant to be merged.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -8,8 +8,7 @@ $finder = PhpCsFixer\Finder::create()
 return PhpCsFixer\Config::create()
     ->setRules([
         '@Symfony' => true,
-        // not sure what should be the equivalent
-        //'empty_return' => false,
+        'simplified_null_return' => false,
         'ordered_imports' => true,
         'return_type_declaration' => ['space_before' => 'one'],
     ])
--- a/.php_cs
+++ b/.php_cs
@@ -1,12 +1,17 @@
 <?php
 
-$finder = Symfony\CS\Finder\DefaultFinder::create()
+$finder = PhpCsFixer\Finder::create()
     ->exclude('var')
-    ->name('console')
+    ->in(__DIR__)
 ;
 
-return Symfony\CS\Config\Config::create()
-    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
-    ->fixers(['-empty_return', 'ordered_use'])
-    ->finder($finder)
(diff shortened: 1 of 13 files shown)


Dockerfile.ci
```diff
@@ -1,10 +1,11 @@
 ARG commit=develop
-FROM elifesciences/proofreader-php:0.1 AS proofreader
+FROM elifesciences/proofreader-php:0.2 AS proofreader
 FROM elifesciences/annotations_cli:${commit}
 ENV DEBUG=1
 
 USER root
-RUN mkdir -p build && chown www-data:www-data build
+RUN mkdir -p build && chown www-data:www-data build && \
+    touch .php_cs.cache && chown www-data:www-data .php_cs.cache
```
giorgiosironi: this is a default, but it can be overridden with a volume if needed

giorgiosironi: (so I'll separate that in a follow-up PR as there may be some decisions to take like where to put it on the host and in the container)

giorgiosironi:
Ready for a new review, this time it's green.

.php_cs
```diff
@@ -1,12 +1,16 @@
 <?php
 
-$finder = Symfony\CS\Finder\DefaultFinder::create()
+$finder = PhpCsFixer\Finder::create()
     ->exclude('var')
-    ->name('console')
```
thewilkybarkid: Means it no longer checks the `bin/console` file?

giorgiosironi: The finder find it by adding `name()`, but the fixer ignores it nonetheless. Investigating.

.php_cs
```diff
@@ -1,12 +1,17 @@
 <?php
 
-$finder = Symfony\CS\Finder\DefaultFinder::create()
+$finder = PhpCsFixer\Finder::create()
     ->exclude('var')
     ->name('console')
```
giorgiosironi: tested `console` and it's now working

PR #81, opened 2018-04-13
[WIP] ELPP-3588 Separate Docker build step

thewilkybarkid:
Also try and add some optimisations.


.dockerignore
```diff
@@ -1,7 +1,11 @@
 build/
 config.php
+config.php.example
 docker-compose.yml
 .git
+.gitignore
 Jenkinsfile
+Jenkinsfile.*
```
giorgiosironi: Just `Jenkinsfile*` is probably enough

Jenkinsfile
```diff
@@ -11,18 +11,18 @@ elifePipeline {
...
             stage 'Build images', {
                 checkout scm
-                sh "IMAGE_TAG=${commit} docker-compose -f docker-compose.ci.yml build"
+                sh "IMAGE_TAG=${commit} COMPOSER_TYPE=no-dev docker-compose build"
             }
 
             stage 'Project tests', {
                 try {
-                    sh "chmod 777 build/ && IMAGE_TAG=${commit} docker-compose -f docker-compose.ci.yml run --rm ci ./project_tests.sh"
+                    sh "chmod 777 build/ && IMAGE_TAG=${commit} docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm ci ./project_tests.sh"
```
giorgiosironi: isn't the double `-f` with both the `dev` and `ci` files necessary everywhere `docker-compose` is called in the build?

docker-compose.yml
```diff
@@ -1,52 +1,74 @@
...
             dockerfile: Dockerfile.cli
-        image: elifesciences/annotations_cli
+            args:
+                composer_type: ${COMPOSER_TYPE}
+                image_tag: ${IMAGE_TAG}
+                dependencies_api_dummy: "${DEPENDENCIES_API_DUMMY}"
+                dependencies_hypothesis_dummy: "${DEPENDENCIES_HYPOTHESIS_DUMMY}"
+        image: elifesciences/annotations_cli:${IMAGE_TAG}
         volumes:
-            - ./:/srv/annotations
```
giorgiosironi: the mounting of the source folder as a volume is one of the main divergences between `dev` and `ci`. Using this file in `ci` correctly avoids mounting that folder (otherwise the content of the image is overwritten and the tests do not target it); but I thought local development needed these mounts

Dockerfile.composer.no-dev
```diff
@@ -0,0 +1,12 @@
+FROM composer:1.6.3
+
+COPY composer.json \
+    composer.lock \
+    ./
+
+RUN composer --no-interaction install --no-dev --ignore-platform-reqs --no-autoloader --no-scripts --no-suggest
```
giorgiosironi: `--ignore-platform-reqs` is what is key here, because it's what normally prevents using a generic image to run composer commands. I assume a missing extension will be caught when tests are run in the image.

Dockerfile.fpm
```diff
@@ -1,27 +1,23 @@
...
+WORKDIR ${PROJECT_FOLDER}
+RUN mkdir -p var/logs var/cache/html_purifier && \
+    chown --recursive elife:elife . && \
+    chown www-data:www-data var/logs var/cache/html_purifier
+
+COPY --chown=elife:elife smoke_tests_fpm.sh .
+COPY --chown=elife:elife web/ web/
+COPY --chown=elife:elife config/ config/
+COPY --from=build --chown=elife:elife /app/vendor/ vendor/
+COPY --from=build --chown=elife:elife /app/src/ src/
```
giorgiosironi: some symlinking necessary for backward compatibility? https://github.com/elifesciences/annotations-formula/blob/cf123e78b2f9eb020759bb4ec865353550cbc10e/salt/annotations/config/home-deployuser-annotations-docker-compose.yml#L22-L23

Dockerfile.composer.no-dev
```diff
@@ -0,0 +1,12 @@
...
+
+COPY src/ src/
+
+RUN composer --no-interaction dump-autoload --no-dev --classmap-authoritative
+RUN composer --no-interaction run-script post-install-cmd
```
giorgiosironi: We may need a base image for the `build` project images, since maintaining all the `composer` commands in every project is starting to sound onerous?

giorgiosironi: While this Dockerfile will/should probably be identical in all PHP (micro)services. The only form of reuse seen so far is to generate it, which is automated copy and paste (e.g. `make ...` that writes it down in the project folder).

docker-compose.yml
thewilkybarkid: Need to look at local development, especially with Composer.

Dockerfile.fpm
thewilkybarkid: It's still `/srv/annotations`?

Dockerfile.composer.no-dev
thewilkybarkid: Had wondered about extending the Composer image with our own, that contains the scripts we currently have. But don't know if it's really worth it.

Dockerfile.ci
```diff
@@ -1,20 +1,30 @@
...
 
 USER root
-RUN mkdir -p build && chown www-data:www-data build && \
-    touch .php_cs.cache && chown www-data:www-data .php_cs.cache
+RUN docker-php-ext-install pcntl
+
+WORKDIR ${PROJECT_FOLDER}
+RUN mkdir -p build var/logs var/cache/html_purifier && \
+    touch .php_cs.cache && \
+    chown --recursive elife:elife . && \
```
giorgiosironi: (`--recursive`) shouldn't be needed? Where files are created that are not owned by `elife`?

Dockerfile.composer
```diff
@@ -0,0 +1,16 @@
...
+ARG composer_dev
+
+COPY composer.json \
+    composer.lock \
+    ./
+
+RUN echo "composer --no-interaction install ${composer_dev} --ignore-platform-reqs --no-autoloader --no-scripts --no-suggest"
+RUN composer --no-interaction install ${composer_dev} --ignore-platform-reqs --no-autoloader --no-scripts --no-suggest
+
+COPY tests/ tests/
```
giorgiosironi: is there a reason for the order of `COPY` statements?

docker-compose.override.yml
```diff
@@ -0,0 +1,27 @@
+version: '3'
```
giorgiosironi: isn't this always applied if existing? So would also be read in `ci`?

Dockerfile.composer
thewilkybarkid: Hypothesis that `src` is more likely to change. But 🤷‍♂️.

docker-compose.override.yml
thewilkybarkid: Not when using `-f`.

.env
```diff
@@ -1,4 +1,5 @@
 COMPOSE_PROJECT_NAME=annotations
+COMPOSER_DEV=
```
giorgiosironi: from its name, looks like a boolean argument, but it's options; could be `COMPOSER_ARGS` (similarly for the Dockerfile argument naming)

docker-compose.override.yml
giorgiosironi: Reasoning on how the 3 files are intended to match an environment:
- `dev`: `docker-compose.yml`, `docker-compose.override.yml`
- `ci`: `docker-compose.yml`, `docker-compose.ci.yml`
- demoing the application: just `docker-compose.yml`

docker-compose.ci.yml
```diff
@@ -1,86 +1,35 @@
...
-            - ./web:/srv/annotations/web
-        depends_on:
-            - fpm
-    fpm:
-        build: 
+    composer:
+        build:
+            args:
+                composer_dev: --no-dev
+    composer-dev:
```
giorgiosironi: I'm not sure how far `docker-compose.yml` and `docker-compose.ci.yml` are from unification right now, but maybe we should put both these `composer` images in `docker-compose.yml` to have a complete stack there. If I want to run tests locally I shouldn't need to use `docker-compose.ci.yml`?

.env
thewilkybarkid: Originally tried to make it a boolean, but `--dev` doesn't exist on `dump-autoload` (it's deprecated on `install`). `args` sound ok, has to apply to both commands though (unless we separate them).

docker-compose.ci.yml
thewilkybarkid: Separate them to avoid a double-Composer build in dev. PHPUnit works using `cli`, but `project_tests.sh` won't because of Proofreader.

Dockerfile.composer
```diff
@@ -0,0 +1,16 @@
+FROM composer:1.6.3
+
+ARG composer_dev
```
giorgiosironi: No problem with the argument names, but a slightly higher level of abstraction for the image name like [`build`](https://medium.com/@alexeiled/docker-pattern-the-build-container-b0d0e86ad601) and `build-dev` would be language-agnostic (or `php-build`|`php-build-dev` for a language-specific version). Other projects could easily require `node` or other specific tools to be installed in their own `build` image.

Later on when this pattern is replicated in `journal` (or even just the dummies), it rises the question on how to remove that duplication for ease of maintenance. There are several possible roads here...

thewilkybarkid: I think it'd be useful to isolate each build tool, Journal's going to need both Gulp and Composer.

thewilkybarkid:
Green, but not actually finding the tests output.

giorgiosironi:
Seems to be fine now https://alfred.elifesciences.org/job/pull-requests-projects/job/annotations/job/PR-81/20/testReport/(root)/

Jenkinsfile
```diff
@@ -11,18 +11,23 @@ elifePipeline {
...
-                sh "IMAGE_TAG=${commit} docker-compose -f docker-compose.ci.yml build"
+                sh "IMAGE_TAG=${commit} docker-compose -f docker-compose.yml -f docker-compose.ci.yml build"
             }
 
             stage 'Project tests', {
                 try {
-                    sh "chmod 777 build/ && IMAGE_TAG=${commit} docker-compose -f docker-compose.ci.yml run --rm ci ./project_tests.sh"
+                    sh "docker run --name annotations_tests_${commit} elifesciences/annotations_ci:${commit}"
+                    sh "docker cp annotations_tests_${commit}:/srv/annotations/build/. build"
                     step([$class: "JUnitResultArchiver", testResults: 'build/phpunit.xml'])
```
giorgiosironi: `cp` and archiving of results needs to be in `finally` to always be run, in order to be executed also on the failure of the tests
Can https://github.com/elifesciences/elife-jenkins-workflow-libs/blob/master/vars/dockerComposeProjectTests.groovy or https://github.com/elifesciences/elife-jenkins-workflow-libs/blob/master/vars/dockerProjectTests.groovy be used here?

giorgiosironi:
Making a checklist here too.

giorgiosironi:
For merge:
- [ ] Rename to `COMPOSER_DEV_ARG` and `composer_dev_arg` (to avoid overgeneric `COMPOSER_ARGS`)
- [ ] Check `chown --recursive` is needed
- [ ] Symlink or similar solution has to make something appear in `/srv/annotations` for backward compatibility with the current formula's `docker-compose` volumes
- [ ] `JUnitResultArchiver` should be in `finally` to collect failed tests

Stuff for later:
- possible creation of some tooling to maintain Dockerfiles across different projects (generation/templating/etc.)
- Jenkinsfile extraction of steps

thewilkybarkid:
Not sure what the problem is with `/srv/annotations`? It's not been moved.

thewilkybarkid:
Removing the `chown --recursive` means that `/srv/annotations`, `/srv/annotations/var` and `/srv/annotations/var/cache` are owned by `root`.  Probably not a problem?

giorgiosironi:
Good to use `--recursive` if it means consistent ownership.
I saw `/app` but it is in fact not used in the main image, just in the builders.

PR #82, opened 2018-04-20
Fix New Relic label

thewilkybarkid:


thewilkybarkid:
No, the only change here is https://github.com/elifesciences/elife-base-images/pull/19 which fixes the NR config.

PR #85, opened 2018-05-21
Improve Composer install

thewilkybarkid:
Downloading repos rather than cloning saves ~12M. Also removes `--no-scripts` since I don't think we actually need it.


thewilkybarkid:
(We started the separate scripts execution as Journal used to use it. But I think that's not good practice with Docker, and has been removed there.)

thewilkybarkid:
Only 2 MB down, but it all helps.

PR #88, opened 2018-07-13
Declare hidden dependencies

thewilkybarkid:


src/HypothesisClient/Result/HttpResult.php
```diff
@@ -21,9 +21,6 @@ private function __construct(ArrayResult $result, ResponseInterface $response)
     public static function fromResponse(ResponseInterface $response) : Result
     {
         $data = json_decode($response->getBody(), true);
```
giorgiosironi: does this throw automatically?

giorgiosironi: the tests answer this: yes

composer.json
```diff
@@ -37,6 +48,7 @@
         "elife/api": "dev-master",
         "elife/api-validator": "^1.0@dev",
         "giorgiosironi/eris": "^0.9.0",
+        "justinrainbow/json-schema": "^5.2",
```
giorgiosironi: I thought this would be abstracted away by `elife/api-validator`, but is in fact instantiated directly so it's a real dependency https://github.com/elifesciences/annotations/blob/161b204c516a57535c38e01d3c1dd0af7da8b644/src/Annotations/AppKernel.php#L44

PR #90, opened 2018-07-30
If upsert client failures, log error but don't repeat

nlisgo:


tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -116,6 +119,70 @@ public function it_retries_failures()
...
+            ['Authorization' => $this->authorization, 'User-Agent' => 'HypothesisClient'],
+            json_encode($patch_data)
+        );
+        $patch_response_mess = json_encode(['status' => 'failure', 'reason' => 'Either the resource you requested doesn\'t exist, or you are not currently authorized to see it.']);
+        $patch_response = new Response(404, [], $patch_response_mess);
+        $rejected_patch_response = new RejectedPromise(new BadResponse($patch_response_mess, $patch_request, $patch_response));
+        $this->httpClient
+            ->expects($this->at(1))
+            ->method('send')
+            ->with(RequestConstraint::equalTo($patch_request))
```
giorgiosironi: When testing for a particular response from the stub, you don't need to use `with()` constraints as the focus is on what we do after that response rather than in what we pass. The `at()` constraints are enough for the stub to discriminate between the first and the second call.

tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -116,6 +119,70 @@ public function it_retries_failures()
...
+            ['Authorization' => $this->authorization, 'User-Agent' => 'HypothesisClient'],
+            json_encode($data)
+        );
+        $post_response_mess = json_encode(['status' => 'failure', 'reason' => 'user with username username already exists.']);
+        $post_response = new Response(400, [], $post_response_mess);
+        $rejected_post_response = new RejectedPromise(new BadResponse($post_response_mess, $post_request, $post_response));
+        $this->httpClient
+            ->expects($this->at(0))
+            ->method('send')
+            ->with(RequestConstraint::equalTo($post_request))
```
giorgiosironi: also this `with()` can go, see other comment

tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -116,6 +119,70 @@ public function it_retries_failures()
...
+        );
+        $patch_response_mess = json_encode(['status' => 'failure', 'reason' => 'Either the resource you requested doesn\'t exist, or you are not currently authorized to see it.']);
+        $patch_response = new Response(404, [], $patch_response_mess);
+        $rejected_patch_response = new RejectedPromise(new BadResponse($patch_response_mess, $patch_request, $patch_response));
+        $this->httpClient
+            ->expects($this->at(1))
+            ->method('send')
+            ->with(RequestConstraint::equalTo($patch_request))
+            ->willReturn($rejected_patch_response);
+        $this->prepareCommandTester();
```
giorgiosironi: lines from `$this->prepareCommandTester();` to `$this->assertSame(0, $this->queue->count());` can probably be extracted into `$this->executeItemFromQueue($item)` to keep the test method short

tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -116,6 +119,70 @@ public function it_retries_failures()
...
+        $data = [
+            'authority' => $this->authority,
+            'username' => 'username',
+            'email' => 'username@blackhole.elifesciences.org',
+            'display_name' => 'PreferredName',
+        ];
+        $this->transformer
+            ->expects($this->once())
+            ->method('transform')
+            ->with($item)
```
giorgiosironi: same here for `with()`, incidentally some variables that are only used in `with()` could then be deleted for a shorter and to-the-point test

src/Annotations/Command/QueueWatchCommand.php
```diff
@@ -66,8 +68,14 @@ protected function process(InputInterface $input, QueueItem $item, $entity = nul
...
-            $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
-            $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
+            try {
+                $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
+                $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
+            } catch (BadResponse $e) {
+                // If upsert failures then log error but don't repeat.
+                $this->queue->commit($item);
+                $this->logger->error(sprintf('Hypothesis user "%s" upsert failure.', $user->getUsername()));
+            }
```
giorgiosironi: just confirming that the parent class https://github.com/elifesciences/bus-sdk-php/blob/8e8a21ba42b8ad367e84877236109c440b6c9615/src/Command/QueueCommand.php#L97-L120 doesn't have support for committing `$item` in case of an exception, it always releases it. So this is the right place to do it, then we can pull up the logic in the library later to make it reusable.

tests/Annotations/Command/QueueWatchCommandTest.php
```diff
@@ -116,6 +119,63 @@ public function it_retries_failures()
...
+        $rejected_post_response = new RejectedPromise(new BadResponse($post_response_mess, $post_request, $post_response));
+        $this->httpClient
+            ->expects($this->at(0))
+            ->method('send')
+            ->willReturn($rejected_post_response);
+        $patch_data = [
+            'email' => 'username@blackhole.elifesciences.org',
+            'display_name' => 'PreferredName',
+        ];
+        $patch_request = new Request(
```
giorgiosironi: I understand the fidelity to what happens in reality, but the `Request` objects could easily be dummies (containing for example just the `PATCH`|`POST` value) in this test since we don't read them.

src/Annotations/Command/QueueWatchCommand.php
```diff
@@ -66,8 +67,14 @@ protected function process(InputInterface $input, QueueItem $item, $entity = nul
...
             $user = new User($id, $email, $sanitized_display_name);
-            $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
-            $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
+            try {
+                $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
+                $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
+            } catch (BadResponse $e) {
+                // If upsert failures then log error but don't repeat.
+                $this->queue->commit($item);
+                $this->logger->error(sprintf('Hypothesis user "%s" upsert failure.', $user->getUsername()));
```
giorgiosironi: either log `$e` here, or rethrow it after this line to allow logging from the parent class

giorgiosironi: Correcting myself: do not rethrow because the parent class will attempt to release `$item`. Log `$e` and then this code is done with it

giorgiosironi: IIRC the logging method accepts the `$context` 2nd argument as `['exception' => $e]`

nlisgo:
@giorgiosironi I have reduced the test significantly now. Thanks for the feedback.

src/Annotations/Command/QueueWatchCommand.php
```diff
@@ -66,8 +67,14 @@ protected function process(InputInterface $input, QueueItem $item, $entity = nul
                 $sanitized_display_name = $display_name;
             }
             $user = new User($id, $email, $sanitized_display_name);
-            $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
-            $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
+            try {
+                $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
+                $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
+            } catch (BadResponse $e) {
+                // If upsert fails, log error but don't repeat.
```
thewilkybarkid: Probably only want this on certain status codes (eg 400s)?

nlisgo: I have added a check in the `catch` statement targetting responses `400` and `404`. Any other response and it will reattempt.

nlisgo:
@thewilkybarkid are you happy with the recent changes?

src/Annotations/Command/QueueWatchCommand.php
thewilkybarkid: I'd suggest not retrying any `4xx` response. It's a client error, so I wouldn't expect a retry to work.

src/Annotations/Command/QueueWatchCommand.php
```diff
@@ -72,8 +72,8 @@ protected function process(InputInterface $input, QueueItem $item, $entity = nul
                 $upsert = $this->hypothesisSdk->users()->upsert($user)->wait();
                 $this->logger->info(sprintf('Hypothesis user "%s" successfully %s.', $upsert->getUsername(), ($upsert->isNew() ? 'created' : 'updated')));
             } catch (BadResponse $e) {
-                // If upsert fails, log error but don't repeat.
-                if (in_array($e->getResponse()->getStatusCode(), [Response::HTTP_BAD_REQUEST, Response::HTTP_NOT_FOUND])) {
+                // If client error detected, log error and don't repeat.
+                if (in_array($e->getResponse()->getStatusCode(), range(Response::HTTP_BAD_REQUEST, 499))) {
```
thewilkybarkid: Rather verbose. `$e->getResponse()->getStatusCode() < 500` would be fine.

nlisgo: Done.

PR #92, opened 2018-08-14
Treat annotations of whitespace like page notes

thewilkybarkid:
Workaround for what seems a bug in Hypothesis. Fixes https://github.com/elifesciences/issues/issues/4393.


thewilkybarkid:
Think I've misunderstood the Hypothesis schema. Going to create the test case based on the real example.

PR #93, opened 2018-09-26
Make name truncation Unicode-aware

thewilkybarkid:
A Hypothesis user is currently failing to be created as their name is 30 characters long, but 33 bytes. Currently it's truncating at 30 bytes rather than 30 characters. Character 28 is bytes 30 and 31, so the truncation results in a broken character and `json_encode()` later fails since it's not valid UTF-8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     "require": {
         "php": "^7.0",
         "ext-curl": "*",
+        "ext-mbstring": "*",
         "aws/aws-sdk-php": "^3.38",
         "beberlei/assert": "^2.7",
         "elife/api-client": "^1.0@dev",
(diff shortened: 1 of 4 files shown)


thewilkybarkid:
> the char limit is a visual limit, not related to any database table

Hypothesis docs says 30 characters. I hope they actually mean characters.

giorgiosironi:
So there is a database underneath having some limit. Checking where this value ends up.

giorgiosironi:
Just made a test using https://hypothes.is/account/profile with this display name:
```
$ php -r "var_dump(strlen('教育漢字教育漢字教育漢字'));"
int(36)
$ php -r "var_dump(mb_strlen('教育漢字教育漢字教育漢字'));"
int(12)
```
was able to save it, but not `1234567890123456789012345678901` for example; the limit truly is 30 characters.

giorgiosironi:
There is also another 30-character limit (and more constraints) on `username`, which here is `$id` so doesn't have problems.

